PR #2, opened 2024-02-18
Add FileStorage struct

twitu:
FileStorage struct retains core functionality from [Kotlin implementation](https://github.com/ARK-Builders/arklib-android/blob/main/lib/src/main/java/dev/arkbuilders/arklib/data/storage/FileStorage.kt#L35).

* Read data from file into a mapping from `ResourceId` to value
* Write data mapping from `ResourceId` to value to file
* Verify file version
* Delete file when dropped

### DIFF
--- a/src/storage/file_storage.rs
+++ b/src/storage/file_storage.rs
@@ -1,0 +1,176 @@
+use std::fs::{self, File};
+use std::io::{BufRead, BufReader, BufWriter, Write};
+use std::str::FromStr;
+use std::time::SystemTime;
+use std::{
+    collections::HashMap,
+    fmt::Display,
+    path::{Path, PathBuf},
+};
+
+use crate::id::ResourceId;
+use crate::{ArklibError, Result};
+
+const LOG_PREFIX: &str = "[file-storage]";
+const STORAGE_VERSION: i32 = 2;
+const STORAGE_VERSION_PREFIX: &str = "version ";
+const KEY_VALUE_SEPARATOR: char = ':';
+
+pub struct FileStorage {
+    log_prefix: String,
+    label: String,
+    path: PathBuf,
+    timestamp: SystemTime,
+}
+
+impl FileStorage {
+    /// Create a new file storage with a diagnostic label and file path
+    pub fn new(label: String, path: &Path) -> Self {
+        Self {
+            label: label.clone(),
+            log_prefix: format!("{} {}", LOG_PREFIX, label).to_string(),
+            path: PathBuf::from(path),
+            timestamp: SystemTime::now(),
+        }
+    }
+
+    /// Read data from disk
+    ///
+    /// Data is read as a key value pairs separated by a symbol and stored
+    /// in a [HashMap] with [ResourceId] key and a generic V value. A handler
+    /// is called on the data after reading it.
+    pub fn read_from_disk<V>(
+        &self,
+        handle: impl FnOnce(HashMap<ResourceId, V>) -> (),
+    ) -> Result<()>
+    where
+        V: FromStr,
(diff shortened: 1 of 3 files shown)


LICENSE
```diff
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2022 ARK Builders
+Copyright (c) 2024 ARK Builders
```
kirillt, line 3: Should it be the latest or the earliest year of activity of the license?

fonts/fonts.json
```diff
@@ -0,0 +1,39 @@
+{
```
kirillt, line 1: Does this folder come from PDF functionality?

src/storage/file_storage.rs
```diff
@@ -0,0 +1,176 @@
...
+                    if line.is_empty() {
+                        continue;
+                    }
+
+                    let parts: Vec<&str> =
+                        line.split(KEY_VALUE_SEPARATOR).collect();
+                    let id = ResourceId::from_str(parts[0])?;
+                    let value = V::from_str(parts[1])?;
+
+                    // TODO: neutral value check when introducing monoids
```
kirillt, line 83: TODOs should be resolved before merging.

src/storage/file_storage.rs
```diff
@@ -0,0 +1,176 @@
...
+        let file = File::create(&self.path)?;
+        let mut writer = BufWriter::new(file);
+
+        writer.write_all(
+            format!("{}{}\n", STORAGE_VERSION_PREFIX, STORAGE_VERSION)
+                .as_bytes(),
+        )?;
+
+        for (id, value) in value_by_id {
+            // TODO: Is neutral check might be needed when monoids are introduced
```
kirillt, line 119: TODOs should be resolved before merging.

src/storage/file_storage.rs
twitu: https://github.com/ARK-Builders/arklib-android/blob/377734870453889f3e9a8821bb1cac3f7f299ab8/lib/src/main/java/dev/arkbuilders/arklib/data/metadata/MetadataStorage.kt#L18

`isNeutral` is defined here in arklib-android repo. I'm not sure how it's related to file storage and why it's a constant.

src/storage/file_storage.rs
twitu: Same as previous comment about `isNeutral` function

src/storage/file_storage.rs
kirillt: This one in `MetadataStorage` is only one of the cases. It's defined as `false` because monoid-based merge strategy isn't used there. See `MonoidIsNotUsed` in the type header. It isn't used because it's cache and we always want to use it locally. That's one of those moments when you implement a universal approach and then one thing doesn't fit nicely into it...

In storages where monoid-based merge-strategy is used, `isNeutral` means same as `== 0`, or `== emptySet()` etc. Depends on what kind of monoid is used.

Ideally, we want to use something more flexible than Monoid which can only sum things up. Implementing CRDT would be great, but we can leave it for next PRs.

fonts/fonts.json
twitu: Removed these files.

LICENSE
twitu: I guess the earliest year of activity is fine - https://softwareengineering.stackexchange.com/questions/210472/is-renewal-of-mit-license-needed-on-github-at-the-beginning-of-each-year

src/storage/file_storage.rs
twitu: Ok, I'll remove the TODOs, We'll take this up as a separate issue when we look into merging data.

PR #5, opened 2024-02-21
Refactor Arklib into Separate Crates

tareknaser:
This PR refactors Arklib into three separate crates for better organization and modularity:

1. `fs-atomic-versions`: Manages the `app_id` file
2. `fs-index`: Handles resource index management
3. `fs-utils`: Contains utility functions and common code used across the project.

Additionally, this PR relocates the `README.md`, `fonts/`, and `testdata/` directories to the project's root directory

- [x] Change target of the PR from `monorepo` to `main` before merging


tareknaser:
Note: `Run benchmarks` check is failing because the main branch still doesn't have benchmarks

data-resource/src/lib.rs
```diff
@@ -0,0 +1,3 @@
+mod id;
```
kirillt, line 1: Probably we can just use `lib.rs` until we'll really have something else except ids.

fs-index/src/pdf.rs
```diff
@@ -0,0 +1,81 @@
+use std::{
+    io::{Read, Seek},
+    path::PathBuf,
+};
+
+use image::DynamicImage;
+use once_cell::sync::OnceCell;
+use pdfium_render::prelude::*;
+
+static PDFIUM: OnceCell<Pdfium> = OnceCell::new(); // static initializers must impl Sync + Send
```
kirillt, line 10: PDF-related code should be extracted into separate crate `data-pdf`.

fs-index/src/link.rs
```diff
@@ -0,0 +1,341 @@
...
+use scraper::{Html, Selector};
+use serde::{Deserialize, Serialize};
+use std::fmt;
+use std::path::Path;
+use std::str::{self, FromStr};
+use std::{io::Write, path::PathBuf};
+use url::Url;
+
+#[derive(Debug, Deserialize, Serialize)]
+pub struct Link {
```
kirillt, line 18: Links-related code should be extracted into `data-link`.

fs-index/src/lib.rs
```diff
@@ -3,39 +3,29 @@
...
 
-pub mod app_id;
-pub mod id;
 pub mod index;
-
 pub mod link;
 pub mod pdf;
 
-mod atomic;
 mod storage;
```
kirillt, line 12: Storages should be defined in their own crate, too: `fs-storage`

FS because they could be implemented with some other backend later.

fs-utils/src/lib.rs
```diff
@@ -0,0 +1,2 @@
+pub mod errors;
+pub mod json;
```
kirillt, line 2: Let's extract these two into `data-error` and `data-json`.

data-resource/src/lib.rs
tareknaser: Sure!
I also think we can port https://github.com/ARK-Builders/arklib/pull/90 to `ark-rust` when we have this PR merged

kirillt (approved):
Thanks @tareknaser

Issue #1, opened 2024-01-29
`fs-storage`: Implement storages in Rust


kirillt:
Right now, we have storages implemented in Kotlin. It would be great to move them from `arklib-android` repo to `arklib` and implement them in Rust. Then, we'll replace Kotlin code with JNI bindings to Rust code.

The original storages implementation is [here](https://github.com/ARK-Builders/arklib-android/tree/main/lib/src/main/java/dev/arkbuilders/arklib/data/storage). Any storage is just a KV-table. Usually, from with `ResourceId` as type for keys.

The original implementation contains also "**sync with disk**" functionality. We should carefully plan this feature. Probably, it should be separated from storages into _merge strategy_ for _atomic files_
* Skip the `Monoid` abstraction for this moment.
* Skip `AggregatedStorage` class for this moment.

- [x] `FileStorage` for keeping the table in a single file 
- [x] `FolderStorage` for keeping the values in separate files
- [ ] `ChunkedStorage` as hybrid between the two, multiple files where each file contains several entries from the table

* What exactly should we version using _atomic files_? Each value, each chunk or the whole table only?

kirillt:
@twitu as soon as the simplest `FileStorage` abstraction is ready, we can start using it in particular implementations like `TagStorage` and `ScoreStorage` and build Tauri components on top of them.

twitu:
Added FileStorage struct with same functionality as [Kotlin implementation](https://github.com/ARK-Builders/arklib-android/blob/main/lib/src/main/java/dev/arkbuilders/arklib/data/storage/FileStorage.kt#L35) in [889638f](https://github.com/ARK-Builders/ark-rust/commit/889638fc98888f26791e670eb64e9cbbfce10f3b)

* Read data from file into a mapping from `ResourceId` to value
* Write data mapping from `ResourceId` to value to file
* Verify file version
* Delete file when dropped

twitu:
[...]

PR #10, opened 2024-03-07
#1: Add FileStorage struct

twitu:
FileStorage struct retains core functionality from [Kotlin implementation](https://github.com/ARK-Builders/arklib-android/blob/main/lib/src/main/java/dev/arkbuilders/arklib/data/storage/FileStorage.kt#L35). #1 

* Read data from file into a mapping from ResourceId to value
* Write data mapping from generic key value pair to file
* Verify file version
* Delete file when dropped

### DIFF
--- a/data-error/src/lib.rs
+++ b/data-error/src/lib.rs
@@ -1,4 +1,4 @@
-use std::str::Utf8Error;
+use std::{convert::Infallible, str::Utf8Error};
 use thiserror::Error;
 
 pub type Result<T> = std::result::Result<T, ArklibError>;
@@ -15,6 +15,9 @@ pub enum ArklibError {
     Parse,
     #[error("Networking error")]
     Network,
+    /// Storage error shows label and error message
+    #[error("Storage error: {0} {1}")]
+    Storage(String, String),
     #[error(transparent)]
     Other(#[from] anyhow::Error),
 }
@@ -46,5 +49,17 @@ impl From<url::ParseError> for ArklibError {
 impl From<Box<dyn std::error::Error>> for ArklibError {
     fn from(e: Box<dyn std::error::Error>) -> Self {
         Self::Other(anyhow::anyhow!(e.to_string()))
+    }
+}
+
+impl From<&str> for ArklibError {
+    fn from(e: &str) -> Self {
+        Self::Other(anyhow::anyhow!(e.to_string()))
+    }
+}
+
+impl From<Infallible> for ArklibError {
+    fn from(_: Infallible) -> Self {
+        Self::Parse
     }
 }
--- a/fs-storage/Cargo.toml
+++ b/fs-storage/Cargo.toml
@@ -8,4 +8,14 @@ name = "fs_storage"
 crate-type = ["rlib"]
 bench = false
 
+[[example]]
+name = "cli"
+
 [dependencies]
+data-error = { path = "../data-error" }
+log = { version = "0.4.17", features = ["release_max_level_off"] }
+
+[dev-dependencies]
+log = { version = "0.4.17", features = ["release_max_level_off"] }
+tempdir = "0.3.7"
+serde_json = "1.0.82"
--- a/fs-storage/Cargo.toml
+++ b/fs-storage/Cargo.toml
@@ -15,6 +15,11 @@ name = "cli"
 data-error = { path = "../data-error" }
 log = { version = "0.4.17", features = ["release_max_level_off"] }
 
+serde_json = "1.0.82"
+anyhow = "1.0.81"
+itertools = "0.12.1"
+serde = { version = "1.0.138", features = ["derive"] }
+
 [dev-dependencies]
 log = { version = "0.4.17", features = ["release_max_level_off"] }
 tempdir = "0.3.7"
--- a/fs-storage/Cargo.toml
+++ b/fs-storage/Cargo.toml
@@ -14,13 +14,9 @@ name = "cli"
 [dependencies]
 data-error = { path = "../data-error" }
 log = { version = "0.4.17", features = ["release_max_level_off"] }
-
 serde_json = "1.0.82"
-anyhow = "1.0.81"
-itertools = "0.12.1"
 serde = { version = "1.0.138", features = ["derive"] }
 
 [dev-dependencies]
-log = { version = "0.4.17", features = ["release_max_level_off"] }
+anyhow = "1.0.81"
 tempdir = "0.3.7"
-serde_json = "1.0.82"
--- a/fs-storage/src/file_storage.rs
+++ b/fs-storage/src/file_storage.rs
@@ -1,0 +1,223 @@
+use std::fmt::Debug;
+use std::fs::{self, File};
+use std::io::{BufRead, BufReader, BufWriter, Write};
+use std::str::FromStr;
+use std::time::SystemTime;
+use std::{
+    collections::BTreeMap,
+    fmt::Display,
+    path::{Path, PathBuf},
+};
+
+use data_error::{ArklibError, Result};
+
+const STORAGE_VERSION: i32 = 2;
+const STORAGE_VERSION_PREFIX: &str = "version ";
+const KEY_VALUE_SEPARATOR: char = ':';
+
+pub struct FileStorage {
+    label: String,
+    path: PathBuf,
+    timestamp: SystemTime,
+}
+
+impl FileStorage {
+    /// Create a new file storage with a diagnostic label and file path
+    pub fn new(label: String, path: &Path) -> Self {
+        Self {
+            label,
+            path: PathBuf::from(path),
+            timestamp: SystemTime::now(),
+        }
+    }
+
+    /// Check if underlying file has been updated
+    ///
+    /// This check can be used before reading the file.
+    pub fn is_file_updated(&self) -> Result<bool> {
+        let file_timestamp = fs::metadata(&self.path)?.modified()?;
+        Ok(self.timestamp < file_timestamp)
+    }
+
+    /// Read data from disk
+    ///
+    /// Data is read as a key value pairs separated by a symbol and stored
+    /// in a [BTreeMap] with a generic key K and V value. A handler
+    /// is called on the data after reading it.
+    pub fn read_file<K, V>(&mut self) -> Result<BTreeMap<K, V>>
+    where
+        K: FromStr + std::hash::Hash + std::cmp::Eq + Debug + std::cmp::Ord,
+        V: FromStr + Debug,
+        ArklibError: From<<K as FromStr>::Err>,
+        ArklibError: From<<V as FromStr>::Err>,
+    {
+        let file = fs::File::open(&self.path)?;
+        let reader = BufReader::new(file);
+        let mut lines = reader.lines();
+
+        let new_timestamp = fs::metadata(&self.path)?.modified()?;
+        match lines.next() {
+            Some(header) => {
+                let header = header?;
+                self.verify_version(&header)?;
+
+                let mut value_by_id = BTreeMap::new();
+                for line in lines {
+                    let line = line?;
+                    if line.is_empty() {
+                        continue;
+                    }
+
+                    let parts: Vec<&str> =
+                        line.split(KEY_VALUE_SEPARATOR).collect();
+                    let id = K::from_str(parts[0])?;
+                    let value = V::from_str(parts[1])?;
+                    value_by_id.insert(id, value);
+                }
+
+                self.timestamp = new_timestamp;
+                Ok(value_by_id)
+            }
+            None => Err(ArklibError::Storage(
+                self.label.clone(),
+                "Storage file is missing header".to_owned(),
+            )),
+        }
+    }
+
+    /// Write data to file
+    ///
+    /// Data is a key-value mapping between [ResourceId] and a generic Value
+    pub fn write_file<K, V>(
+        &mut self,
+        value_by_id: &BTreeMap<K, V>,
+    ) -> Result<()>
+    where
+        K: Display,
+        V: Display,
+    {
+        fs::create_dir_all(self.path.parent().unwrap())?;
+        let file = File::create(&self.path)?;
+        let mut writer = BufWriter::new(file);
+
+        writer.write_all(
+            format!("{}{}\n", STORAGE_VERSION_PREFIX, STORAGE_VERSION)
+                .as_bytes(),
+        )?;
+
+        for (id, value) in value_by_id {
+            writer.write_all(
+                format!("{}{}{}\n", id, KEY_VALUE_SEPARATOR, value).as_bytes(),
+            )?;
+        }
+
+        let new_timestamp = fs::metadata(&self.path)?.modified()?;
+        if new_timestamp == self.timestamp {
+            return Err("Timestamp didn't update".into());
+        }
+        self.timestamp = new_timestamp;
+
+        log::info!(
+            "{} {} entries has been written",
+            self.label,
+            value_by_id.len()
+        );
+        Ok(())
+    }
+
+    pub fn erase(&self) {
+        if let Err(e) = fs::remove_file(&self.path) {
+            log::error!(
+                "{} Failed to delete file because of error: {}",
+                self.label,
+                e
+            )
+        }
+    }
+
+    /// Verify the version stored in the file header
+    fn verify_version(&self, header: &str) -> Result<()> {
+        if !header.starts_with(STORAGE_VERSION_PREFIX) {
+            return Err(ArklibError::Storage(
+                self.label.clone(),
+                "Unknown storage version prefix".to_owned(),
+            ));
+        }
+
+        let version = header[STORAGE_VERSION_PREFIX.len()..]
+            .parse::<i32>()
+            .map_err(|_err| {
+                <&str as Into<ArklibError>>::into(
+                    "Unable to parse storage version",
+                )
+            })?;
+
+        if version != STORAGE_VERSION {
+            return Err(ArklibError::Storage(
+                self.label.clone(),
+                format!(
+                    "Storage version mismatch: expected {}, found {}",
+                    STORAGE_VERSION, version
+                ),
+            ));
+        }
+
+        Ok(())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use std::collections::BTreeMap;
+    use tempdir::TempDir;
+
+    use crate::file_storage::FileStorage;
+
+    #[test]
+    fn test_file_storage_write_read() {
+        let temp_dir =
+            TempDir::new("tmp").expect("Failed to create temporary directory");
+        let storage_path = temp_dir.path().join("test_storage.txt");
+
+        let mut file_storage =
+            FileStorage::new("TestStorage".to_string(), &storage_path);
+
+        let mut data_to_write = BTreeMap::new();
+        data_to_write.insert("key1".to_string(), "value1".to_string());
+        data_to_write.insert("key2".to_string(), "value2".to_string());
+
+        file_storage
+            .write_file(&data_to_write)
+            .expect("Failed to write data to disk");
+
+        let data_read: BTreeMap<_, _> = file_storage
+            .read_file()
+            .expect("Failed to read data from disk");
+
+        assert_eq!(data_read, data_to_write);
+    }
+
+    #[test]
+    fn test_file_storage_auto_delete() {
+        let temp_dir =
+            TempDir::new("tmp").expect("Failed to create temporary directory");
+        let storage_path = temp_dir.path().join("test_storage.txt");
+
+        let mut file_storage =
+            FileStorage::new("TestStorage".to_string(), &storage_path);
+
+        let mut data_to_write = BTreeMap::new();
+        data_to_write.insert("key1".to_string(), "value1".to_string());
+        data_to_write.insert("key2".to_string(), "value2".to_string());
+
+        file_storage
+            .write_file(&data_to_write)
+            .expect("Failed to write data to disk");
+
+        assert_eq!(storage_path.exists(), true);
+
+        file_storage.erase();
+
+        assert_eq!(storage_path.exists(), false);
+    }
+}
--- a/fs-storage/src/file_storage.rs
+++ b/fs-storage/src/file_storage.rs
@@ -39,7 +39,7 @@ impl FileStorage {
 
     /// Read data from disk
     ///
-    /// Data is read as a key value pairs separated by a symbol and stored
+    /// Data is read as key value pairs separated by a symbol and stored
     /// in a [BTreeMap] with a generic key K and V value. A handler
     /// is called on the data after reading it.
     pub fn read_file<K, V>(&mut self) -> Result<BTreeMap<K, V>>
--- a/fs-storage/src/file_storage.rs
+++ b/fs-storage/src/file_storage.rs
@@ -5,7 +5,6 @@ use std::str::FromStr;
 use std::time::SystemTime;
 use std::{
     collections::BTreeMap,
-    fmt::Display,
     path::{Path, PathBuf},
 };
 
@@ -13,7 +12,6 @@ use data_error::{ArklibError, Result};
 
 const STORAGE_VERSION: i32 = 2;
 const STORAGE_VERSION_PREFIX: &str = "version ";
-const KEY_VALUE_SEPARATOR: char = ':';
 
 pub struct FileStorage {
     label: String,
@@ -46,10 +44,14 @@ impl FileStorage {
     /// is called on the data after reading it.
     pub fn read_file<K, V>(&mut self) -> Result<BTreeMap<K, V>>
     where
-        K: FromStr + std::hash::Hash + std::cmp::Eq + Debug + std::cmp::Ord,
-        V: FromStr + Debug,
+        K: serde::de::DeserializeOwned
+            + FromStr
+            + std::hash::Hash
+            + std::cmp::Eq
+            + Debug
+            + std::cmp::Ord,
+        V: serde::de::DeserializeOwned + Debug,
         ArklibError: From<<K as FromStr>::Err>,
-        ArklibError: From<<V as FromStr>::Err>,
     {
         let file = fs::File::open(&self.path)?;
         let reader = BufReader::new(file);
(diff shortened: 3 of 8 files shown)


twitu:
@Pushkarm029 the changes look good. Can you make a separate PR for this and move the logic into a single file in a bin directory. That way cargo knows it's meant to be compiled as an executable. Something like -

```
fs-storage/
   src/
       bin/
           cli.rs
```

https://doc.rust-lang.org/cargo/guide/project-layout.html

The main argument can choose to read or write based on the pass arguments. You can consider using clap (or similar library) to handle cli arguments.

kirillt:
@twitu it seems that we need to rebase this branch

kirillt:
The `write` sample produces an empty folder:
```
[kirill@lenovo debug]$ ./write /tmp/fs-storage-test
Storage directory created successfully at /tmp/fs-storage-test
Enter a key-value pair (key=value), or enter 'done' to finish:
1=a
Enter a key-value pair (key=value), or enter 'done' to finish:
2=b
Enter a key-value pair (key=value), or enter 'done' to finish:
c=3
Enter a key-value pair (key=value), or enter 'done' to finish:
done
Key-Value Pairs:
1: a
2: b
c: 3
[kirill@lenovo debug]$ ls -lah /tmp/fs-storage-test/
total 0
drwxr-xr-x.  2 kirill kirill  40 Mar 11 10:56 .
drwxrwxrwt. 29 root   root   920 Mar 11 10:56 ..
```

fs-storage/src/file_storage.rs
```diff
@@ -0,0 +1,246 @@
...
+use std::time::SystemTime;
+use std::{
+    collections::HashMap,
+    fmt::Display,
+    path::{Path, PathBuf},
+};
+
+use data_error::{ArklibError, Result};
+
+const LOG_PREFIX: &str = "[file-storage]";
```
kirillt, line 14: This came from Kotlin implementation, but let's remove it here. Rust logs stuff a bit differently.

fs-storage/src/file_storage.rs
```diff
@@ -0,0 +1,246 @@
...
+                self.label.clone(),
+                "Storage format is older than the app".to_owned(),
+            )),
+        }
+    }
+}
+
+impl Drop for FileStorage {
+    fn drop(&mut self) {
+        if let Err(e) = fs::remove_file(&self.path) {
```
kirillt, line 181: Does this erases the file from disk every time we discard the storage object? The idea is to persist the storage on disk between runs.

fs-storage/src/file_storage.rs
```diff
@@ -0,0 +1,246 @@
...
+        }
+
+        log::info!("the file was modified externally, merging");
+
+        let value_by_id = self.read_file_from_disk()?;
+        if !value_by_id.is_empty() {
+            handle(value_by_id);
+        }
+
+        Ok(())
```
kirillt, line 70: It seems that `self.timestamp` is not updated as it's done in Kotlin implementation after each read:
https://github.com/ARK-Builders/arklib-android/blob/main/lib/src/main/java/dev/arkbuilders/arklib/data/storage/FileStorage.kt#L99

fs-storage/src/file_storage.rs
```diff
@@ -0,0 +1,246 @@
...
+
+    /// Read data from disk
+    ///
+    /// Data is read as a key value pairs separated by a symbol and stored
+    /// in a [HashMap] with a generic key K and V value. A handler
+    /// is called on the data after reading it.
+    pub fn read_file<K, V>(
+        &self,
+        mut handle: impl FnMut(HashMap<K, V>),
+    ) -> Result<()>
```
kirillt, line 45: I don't remember exact reason why it was implemented this way in Kotlin, but probably this time we can start with simpler and cleaner API and merge `read_file` into `read_file_from_disk`. So, there should be only function returning a mapping (potentially wrapped into Storage value).

fs-storage/src/file_storage.rs
```diff
@@ -0,0 +1,246 @@
...
+const STORAGE_VERSION: i32 = 2;
+const STORAGE_VERSION_PREFIX: &str = "version ";
+const KEY_VALUE_SEPARATOR: char = ':';
+
+pub struct FileStorage {
+    log_prefix: String,
+    label: String,
+    path: PathBuf,
+    timestamp: SystemTime,
+}
```
kirillt, line 24: What about such API?
```
pub struct Mapping<K,V> {
    timestamp: SystemTime,
    values: HashMap<K, V>
}

pub struct Storage<K,V> {
    label: String,
    path: PathBuf,
    mapping: Option<Mapping<K,V>> // `None` means not loaded yet
}

impl Storage {
...
    fn new(label: String, path: &Path) -> Self { ... }
    fn load(self: Self) -> Self { ... }
    fn update(self: Self) -> Self { ... }
...
}
```

Or something like this:
```
pub struct Mapping<K,V> {
    timestamp: SystemTime,
    values: HashMap<K, V>
}

pub struct Storage<K,V> {
    label: String,
    path: PathBuf,
    mapping: Mapping<K,V>
}

impl Storage {
...
    fn provide(label: String, path: &Path) -> Self { ... }
    fn update(self: Self) -> Self { ... }
...
}
```

fs-storage/src/file_storage.rs
twitu: Aah this comes from the erase function implemented in Kotlin. But I guess instead of deleting the file by itself, it can be exposed as a function that can be called by the user.

data-error/src/lib.rs
```diff
@@ -15,6 +15,8 @@ pub enum ArklibError {
     Parse,
     #[error("Networking error")]
     Network,
+    #[error("Storage error: {0} {1}")]
+    Storage(String, String),
```
tareknaser, line 19: Why do we need 2 strings in this error type?
Could you add a doc comment to explain what they represent?

fs-storage/src/file_storage.rs
```diff
@@ -0,0 +1,246 @@
...
+        V: FromStr + Debug,
+        ArklibError: From<<K as FromStr>::Err>,
+        ArklibError: From<<V as FromStr>::Err>,
+    {
+        let file = fs::File::open(&self.path)?;
+        let reader = BufReader::new(file);
+        let mut lines = reader.lines();
+
+        match lines.next() {
+            Some(Ok(header)) => {
```
tareknaser, line 85:
```suggestion
            Some(header) => {
                let header = header?;
```

fs-storage/src/file_storage.rs
```diff
@@ -0,0 +1,246 @@
...
+        if !header.starts_with(STORAGE_VERSION_PREFIX) {
+            return Err(ArklibError::Storage(
+                self.label.clone(),
+                "Unknown storage version".to_owned(),
+            ));
+        }
+
+        let version = header[STORAGE_VERSION_PREFIX.len()..]
+            .parse::<i32>()
+            .unwrap();
```
tareknaser, line 163: We can return the error here instead of using `unwrap()`

fs-storage/src/file_storage.rs
```diff
@@ -0,0 +1,246 @@
...
+            std::cmp::Ordering::Equal => Ok(()),
+            std::cmp::Ordering::Less => Err(ArklibError::Storage(
+                self.label.clone(),
+                "Storage format is newer than the app".to_owned(),
+            )),
+            std::cmp::Ordering::Greater => Err(ArklibError::Storage(
+                self.label.clone(),
+                "Storage format is older than the app".to_owned(),
+            )),
+        }
```
tareknaser, line 175:
```suggestion
        if version != STORAGE_VERSION {
            return Err(ArklibError::Storage(
                self.label.clone(),
                format!(
                    "Storage version mismatch: expected {}, found {}",
                    STORAGE_VERSION, version
                ),
            ));
        }
        
        Ok(())
```

tareknaser (commented):
Thanks for putting this together

Left some comments

fs-storage/src/cli/read.rs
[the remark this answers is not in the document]
kirillt: @tareknaser that was explicit requirement in the task to implement a tiny self-contained CLI sample to test only this feature. This would allow more versatile testing comparing to unit tests, without making too much effort for creating fully-fledged CLI.

We will check this sample CLI (or maybe better say PoC) against existing storages support in `ark-cli` later and merge if necessary.

fs-storage/src/cli/read.rs
```diff
@@ -0,0 +1,36 @@
+use std::env;
+use std::path::Path;
+use fs_storage::file_storage::FileStorage;
+
+fn main() {
+    let args: Vec<String> = env::args().collect();
+    if args.len() != 3 {
+        println!("Usage: {} <path_to_storage> <key>", args[0]);
+        return;
+    }
```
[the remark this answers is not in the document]
kirillt: Good idea but only if it won't be over-engineering. This sample CLI is allowed to follow strict parameters format and isn't required to e user-friendly (only developers would use it for testing).

fs-storage/src/file_storage.rs
kirillt: >  It'll be helpful to know why it was implemented callback style in Kotlin. So that later the design doesn't cause issue.

Agree, but it seems to be too difficult at this moment.. This time, we'll create core components basing on common sense and then adjust existing apps to them.

fs-storage/src/file_storage.rs
kirillt: Yep, we need erasing function in the API but must not call it automatically.

tareknaser:
If the "Verify build" CI test is blocking this, you can cherry pick ba41bd8690c03920cfaaf0d0f94c13a7342e950d to fix it

fs-storage/src/cli/read.rs
```diff
@@ -0,0 +1,36 @@
...
+            println!("Value for key '{}': {:?}", key, value);
+        } else {
+            println!("Key '{}' not found in storage.", key);
+        }
+    }) {
+        Ok(_) => {}
+        Err(e) => {
+            println!("Error reading storage file: {:?}", e);
+        }
+    }
```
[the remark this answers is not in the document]
twitu: I've rewritten the cli differently

fs-storage/src/cli/write.rs
```diff
@@ -0,0 +1,72 @@
...
+        kv_pairs.insert(key, value);
+    }
+
+    println!("Key-Value Pairs:");
+    for (key, value) in &kv_pairs {
+        println!("{}: {}", key, value);
+    }
+
+    if let Err(err) = write_to_file(kv_pairs, storage_path) {
+        println!("Error writing to file: {}", err);
```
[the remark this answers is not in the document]
twitu: Since this a dev cli errors are unwrapped in the new implementation

fs-storage/src/file_storage.rs
twitu: Fixed the logic to update timestamp from metadata of read file.

fs-storage/src/file_storage.rs
twitu: I've updated the api to remove the callback style and just return the hashmap directly. Currently I don't see how the additional complexity of wrapping the HashMap in a Mapping will help.

I think the current state of the PR is a good first version of a functional file storage implementation. I propose we merge it and add more features as separate issues. I prefer not to keep PRs open for too long because they lead to merge conflicts.

fs-storage/src/bin/cli.rs
```diff
@@ -0,0 +1,64 @@
+use std::collections::HashMap;
+use std::env;
+use std::path::Path;
+
+use fs_storage::file_storage::FileStorage;
+
+fn main() {
+    let args: Vec<String> = env::args().collect();
+    if args.len() < 3 {
+        println!("Usage: cargo run -- [read|write] path [key1,key2|key1:value1,key2:value2]");
```
kirillt: This universal message could be confusing during quick evaluation.

I've committed a simpler instruction with separate commands for read and write.

kirillt:
@twitu The samples work, thanks. I've noticed one more thing to fix:
```
$ cargo run -- write /tmp/x a:1,b:2,c:3,d:4
...
$ cat /tmp/x
version 2
a:1
c:3
b:2
d:4
$ cargo run -- write /tmp/x a:1,b:2,c:3,d:4
...
$ cat /tmp/x
version 2
b:2
a:1
d:4
c:3
```

- [ ] Storages should be persisted deterministically, i.e. the storage files must be equal for equal KV tables. In Rust, iterating through `BTreeMap` is deterministic.

And also https://github.com/ARK-Builders/ark-rust/pull/10#issuecomment-1988678975

fs-storage/src/file_storage.rs
kirillt: Agree about the conflicts. But if the changes are small it's better to make it in same PR to avoid excessive task management.

`Mapping` wrapper is useful only because it also contains timestamp. We could use a tuple, but named structure is better. The rest looks good for now.

kirillt:
I've converted https://github.com/ARK-Builders/ark-rust/pull/10#issuecomment-1988678975 into separate issues:
- #18
- #19

**But let's have deterministic output in this PR.**

Also, it seems that this branch and main have same commits but with different hashes. This causes duplicate commits during rebase. Probably we can squash though.

fs-storage/src/bin/cli.rs
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    let command = &args[1];
+    let path = &args[2];
+
+    match command.as_str() {
+        "read" => {
+            let keys = if args.len() > 3 {
+                args[3]
+                    .split(",")
```
tareknaser, line 23:
```suggestion
                    .split(',')
```

fs-storage/src/bin/cli.rs
```diff
@@ -0,0 +1,64 @@
...
+        }
+        "write" => {
+            if args.len() < 4 {
+                println!(
+                    "Usage: cargo run -- write path key1:value1,key2:value2"
+                );
+                return;
+            }
+            let kv_pairs = args[3]
+                .split(",")
```
tareknaser, line 53:
```suggestion
                .split(',')
```

fs-storage/src/bin/cli.rs
```diff
@@ -0,0 +1,64 @@
...
+                .map(|s| {
+                    let kv: Vec<&str> = s.split(":").collect();
```
tareknaser, line 55:
```suggestion
                    let kv: Vec<&str> = s.split(':').collect();
```

fs-storage/src/bin/cli.rs
```diff
@@ -0,0 +1,64 @@
...
+use std::env;
+use std::path::Path;
+
+use fs_storage::file_storage::FileStorage;
+
+fn main() {
+    let args: Vec<String> = env::args().collect();
+    if args.len() < 3 {
+        println!("Usage:");
+        println!("    cargo run -- write <path> <key1:value1,key2:value2,...>");
```
tareknaser, line 11: Using `cargo run -- write <path> <key1:value1,key2:value2,...>` may become challenging if we need to pass several key-value pairs. 
Maybe providing them in JSON format would be a cleaner approach. What do you think?

tareknaser: It can also be easier to parse and use than splitting each time

fs-storage/src/file_storage.rs
```diff
@@ -0,0 +1,223 @@
...
+            )
+        }
+    }
+
+    /// Verify the version stored in the file header
+    fn verify_version(&self, header: &str) -> Result<()> {
+        if !header.starts_with(STORAGE_VERSION_PREFIX) {
+            return Err(ArklibError::Storage(
+                self.label.clone(),
+                "Unknown storage version".to_owned(),
```
tareknaser, line 143: Maybe the error message should be that the version prefix wasn't found

fs-storage/src/file_storage.rs
```diff
@@ -0,0 +1,223 @@
...
+
+const STORAGE_VERSION: i32 = 2;
+const STORAGE_VERSION_PREFIX: &str = "version ";
+const KEY_VALUE_SEPARATOR: char = ':';
+
+pub struct FileStorage {
+    label: String,
+    path: PathBuf,
+    timestamp: SystemTime,
+}
```
tareknaser, line 22: I am curious what `label` represents here.
I guess also `timestamp` is the latest modified time for the whole `FileStorage`  instance?

tareknaser (approved):
I left some minor comments to make `cargo clippy` happy but LGTM
Thanks

fs-storage/src/bin/cli.rs
Pushkarm029: @tareknaser, yes, it will be better. working on it

fs-storage/src/file_storage.rs
Pushkarm029: @twitu

twitu: `label` is more of telemetry/debugging/logging thing ported from the Kotlin implementations. The logs for file storage operations have the label. It probably helps understanding the logs if an application uses multiple file storage instances.

Yes `timestamp` is the latest modified time for the underlying file. It is updated when the file is written to or when the file is read from.

tareknaser: > The logs for file storage operations have the label. It probably helps understanding the logs if an application uses multiple file storage instances.

makes perfect sense

fs-storage/src/bin/README.md
```diff
@@ -0,0 +1,31 @@
...
+```bash
+cd fs-storage/src/bin
+```
+
+- Run Write Command
+```bash
+cargo run -- write /tmp/z test.json
+```
+
+- Run Read Command
```
tareknaser, line 21: I noticed from the code that if no keys were provided for the `read` command, the whole map gets printed to the terminal

Can you add a note about that?

fs-storage/src/bin/cli.rs
```diff
@@ -0,0 +1,80 @@
...
+            if keys.is_empty() {
+                for (key, value) in map {
+                    println!("{}: {}", key, value);
+                }
+            } else {
+                for key in &keys {
+                    if let Some(value) = map.get(key) {
+                        println!("{}: {}", key, value);
+                    } else {
+                        println!("Key '{}' not found", key);
```
tareknaser, line 41:
```suggestion
                        eprintln!("Key '{}' not found", key);
```

fs-storage/src/bin/cli.rs
```diff
@@ -0,0 +1,80 @@
...
+                }
+            } else {
+                println!("JSON value is not an object");
+                return;
+            }
+
+            let mut fs = FileStorage::new("cli".to_string(), Path::new(path));
+            fs.write_file(&kv_pairs).unwrap();
+        }
+        _ => println!("Invalid command. Use 'read' or 'write'."),
```
tareknaser, line 78:
```suggestion
        _ => eprintln!("Invalid command. Use 'read' or 'write'."),
```

fs-storage/Cargo.toml
```diff
@@ -9,3 +9,17 @@ crate-type = ["rlib"]
...
+data-resource = { path = "../data-resource" }
+log = { version = "0.4.17", features = ["release_max_level_off"] }
+
+serde_json = "1.0.82"
+serde = { version = "1.0.138", features = ["derive"] }
+
+[dev-dependencies]
+log = { version = "0.4.17", features = ["release_max_level_off"] }
+tempdir = "0.3.7"
+tempfile = "3.10.1"
```
tareknaser, line 24: We use `tempdir` in tests for other crates. 
If there is no reason to prefer `tempfile`, we shouldn't use it only in this crate

fs-storage/Cargo.toml
```diff
@@ -9,3 +9,17 @@ crate-type = ["rlib"]
 bench = false
 
 [dependencies]
+data-error = { path = "../data-error" }
+data-json = { path = "../data-json" }
+fs-atomic-versions = { path = "../fs-atomic-versions" }
+data-resource = { path = "../data-resource" }
+log = { version = "0.4.17", features = ["release_max_level_off"] }
+
+serde_json = "1.0.82"
```
tareknaser, line 18: `serde_json` is only needed for the binary. We shouldn't depend on it for the rest of the package. 

I came across a similar question on stackoverflow. Here is [an answer with some suggestions](https://stackoverflow.com/a/46017284)

fs-storage/Cargo.toml
```diff
@@ -9,3 +9,17 @@ crate-type = ["rlib"]
 bench = false
 
 [dependencies]
+data-error = { path = "../data-error" }
+data-json = { path = "../data-json" }
+fs-atomic-versions = { path = "../fs-atomic-versions" }
+data-resource = { path = "../data-resource" }
```
tareknaser, line 15: These 3 crates are never used

fs-storage/Cargo.toml
```diff
@@ -9,3 +9,17 @@ crate-type = ["rlib"]
...
+log = { version = "0.4.17", features = ["release_max_level_off"] }
+
+serde_json = "1.0.82"
+serde = { version = "1.0.138", features = ["derive"] }
```
tareknaser, line 19: `serde` is not used as well

tareknaser (changes requested):
I gave the binary a spin locally and everything works as expected. Thanks for adding support for JSON files!

Left some comments
We also need to make sure the CI is green before merging. Currently the benchmarks one is failing but I added a suggestion to fix it

fs-storage/Cargo.toml
```diff
@@ -9,3 +9,17 @@ crate-type = ["rlib"]
 bench = false
```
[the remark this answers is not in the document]
tareknaser: This should fix the benchmarks CI

fs-storage/src/cli/write.rs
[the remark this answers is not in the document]
tareknaser: > Since this a dev cli errors are unwrapped in the new implementation

I second this.
I know it's a grey area in rust where people take sides in general. For our case (CLI), I think we should never panic

I never want to see this output as a user when running a CLI command
```bash
thread 'main' panicked at fs-storage/src/bin/cli.rs:54:18:
Failed to read JSON file: Os { code: 2, kind: NotFound, message: "No such file or directory" }
```
I prefer passing down the error or using `anyhow::Context` 
Here is a reference from the "Command Line Applications in Rust" book
https://rust-cli.github.io/book/tutorial/errors.html

What do you think?

tareknaser:
Let's also rebase the branch on top of main

rust-toolchain.toml
```diff
@@ -0,0 +1,3 @@
+[toolchain]
+version = "1.75.0"
```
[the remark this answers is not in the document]
tareknaser: A good resource: https://gribnau.dev/cargo-msrv/commands/find.html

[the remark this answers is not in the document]
twitu: The toolchain is what the compiler and tooling for this workspace will default too. Is there any requirement to support older versions? Using older versions can give difficulty later in upgrading libraries. Unless there's some specific requirement to support older versions we should go ahead with the latest and greatest. Especially so because we plan to pull in the tokio and async ecosystem which is still evolving.

fs-storage/Cargo.toml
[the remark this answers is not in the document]
twitu: Made cli an example which should also fix this.

twitu:
Fixed the commit history.

I also experimented with creating a mapping struct in separate [branch](https://github.com/ARK-Builders/ark-rust/tree/file-storage-temp).

```
pub struct Mapping<K, V> {
    timestamp: SystemTime,
    values: HashMap<K, V>,
}
```

It adds extra complexity of generics and also a few unresolved problems around consistency of the timestamp. Basically we have two timestamps in the system now. One for the mapping and one for the file storage struct. This leads to duplication without clear distinction between how they are different or what their relation is to each other.

Secondly, the save function doesn't actually use the timestamp in the mapping since it reads the current time.

```
    /// Write data to file
    ///
    /// Data is a key-value mapping between [ResourceId] and a generic Value
    pub fn save<K, V>(&mut self, value_by_id: Mapping<K, V>) -> Result<()>
```

The mapping struct is only used in one place where it is returned by the `load` function.

```
    fn load<K, V>(&mut self) -> Result<Mapping<K, V>>
```

We can consider adding it later when it has more use cases.

This is a good first version of the FileStorage logic and we can merge it once @Pushkarm029 gives the cli a final refactor.

fs-storage/src/file_storage.rs
```diff
@@ -0,0 +1,224 @@
...
+    ///
+    /// This check can be used before reading the file.
+    pub fn is_file_updated(&self) -> Result<bool> {
+        let file_timestamp = fs::metadata(&self.path)?.modified()?;
+        Ok(self.timestamp < file_timestamp)
+    }
+
+    /// Read data from disk
+    ///
+    /// Data is read as a key value pairs separated by a symbol and stored
```
tareknaser, line 42:
```suggestion
    /// Data is read as key value pairs separated by a symbol and stored
```

fs-storage/src/file_storage.rs
```diff
@@ -0,0 +1,224 @@
...
+                "Unknown storage version prefix".to_owned(),
+            ));
+        }
+
+        let version = header[STORAGE_VERSION_PREFIX.len()..]
+            .parse::<i32>()
+            .map_err(|_err| {
+                <&str as Into<ArklibError>>::into(
+                    "Unable to parse storage version",
+                )
```
tareknaser, line 152:
```suggestion
                ArklibError::Storage(
                    self.label.clone(),
                    "Failed to parse storage version".to_owned(),
                )
```

fs-storage/src/file_storage.rs
```diff
@@ -0,0 +1,224 @@
...
+
+        log::info!(
+            "{} {} entries have been written",
+            self.label,
+            value_by_id.len()
+        );
+        Ok(())
+    }
+
+    pub fn erase(&self) -> Result<()> {
```
tareknaser, line 128: This method deserves a doc comment

fs-storage/src/file_storage.rs
```diff
@@ -0,0 +1,224 @@
...
+        if let Err(e) = fs::remove_file(&self.path) {
+            return Err(ArklibError::Storage(
+                self.label.clone(),
+                format!("Failed to delete file because of error: {}", e),
+            ));
+        }
+        Ok(())
+    }
```
tareknaser, line 136:
```suggestion
    pub fn erase(&self) -> Result<()> {
        fs::remove_file(&self.path).map_err(|err| {
            ArklibError::Storage(self.label.clone(), err.to_string())
        })
    }
```

tareknaser (approved):
Looks good to me!
Thank you

All the comments here are nits. Feel free to ignore them

PR #23, opened 2024-04-08
`fs-storage`: Port BaseStorage abstraction from Kotlin

### LINKED ISSUE #18: `fs-storage`: Port `BaseStorage` abstraction from Kotlin
Port functionality from [BaseStorage.kt](https://github.com/ARK-Builders/arklib-android/blob/main/lib/src/main/java/dev/arkbuilders/arklib/data/storage/BaseStorage.kt) into `fs-storage`.

It seems to me, that it should be a `trait `.

Pushkarm029:
Resolves #18


kirillt:
Good start, let @tareknaser know please, when the todos are solved and ready for review.

fs-storage/src/base_storage.rs
```diff
@@ -0,0 +1,34 @@
...
+    ///
+    /// This check can be used before reading the file.
+    fn is_file_updated(&self) -> Result<bool>;
+
+    /// Read data from disk
+    ///
+    /// Data is read as key value pairs separated by a symbol and stored
+    /// in a [BTreeMap] with a generic key K and V value. A handler
+    /// is called on the data after reading it.
+    fn read_file(&mut self) -> Result<BTreeMap<K, V>>;
```
kirillt, line 28: Should be something without word `file`, probably `read_on_disk`? Or `read_fs`?

Because `FolderStorage` would implemente `BaseStorage` but it reads many files (a folder).

fs-storage/src/base_storage.rs
```diff
@@ -0,0 +1,34 @@
...
+
+    /// Write data to file
+    ///
+    /// Data is a key-value mapping between [ResourceId] and a generic Value
+    fn write_file(&mut self) -> Result<()>;
```
kirillt, line 33: Same here

fs-storage/src/base_storage.rs
```diff
@@ -0,0 +1,34 @@
...
+use std::str::FromStr;
+
+use data_error::Result;
+
+pub trait BaseStorage<K, V>
+where
+    K: FromStr + Hash + Eq + Ord + Debug + Clone,
+    V: Debug + Clone,
+{
+    fn get(&self, id: &K) -> Option<V>;
```
twitu, line 13: Since `BaseStorage` is an abstraction of top of a file/files that store key value mappings. There might be a way to derive [AsRef](https://doc.rust-lang.org/std/convert/trait.AsRef.html) for BaseStorage and to get most immutable BtreeMap related functions for free.

fs-storage/src/file_storage.rs
```diff
@@ -8,51 +9,104 @@ use std::{
...
     label: String,
     path: PathBuf,
     timestamp: SystemTime,
+    pub value_by_id: BTreeMap<K, V>,
 }
 
-impl FileStorage {
+impl<K, V> FileStorage<K, V>
+where
+    K: serde::Serialize,
```
twitu, line 31: Generic parameter conditions are generally not mentioned where the struct is declared. This because it needless prevents constructing the struct. It is preferred to mention these conditions when implementing a trait/function that need the conditions.

fs-storage/src/base_storage.rs
twitu: Notice that we don't want to derive AsRefMut because we have some custom logic to update timestamp when doing mutable operations.

twitu (commented):
The trait implementation looks solid!

fs-storage/src/base_storage.rs
```diff
@@ -0,0 +1,41 @@
...
+    fn set(&mut self, id: K, value: V) -> Result<()>;
+    fn remove(&mut self, id: &K) -> Result<()>;
+
+    /// Remove file at stored path
+    fn erase(&mut self) -> Result<()>;
+
+    /// Get immutable BTreeMap
+    ///
+    /// This can be used to get most immutable BtreeMap related functions for free.
+    fn as_ref(&self) -> &BTreeMap<K, V>;
```
tareknaser, line 23: I am not sure if this is truly necessary.
If `FileStorage` and `FolderStorage` directly manage their own internal storage (e.g., a `BTreeMap`), they might not need to expose it directly.

fs-storage/src/file_storage.rs
tareknaser, line 1: Could you add unit tests for new `FileStorage` methods?

fs-storage/src/base_storage.rs
```diff
@@ -0,0 +1,41 @@
...
+
+    /// Check if underlying file has been updated
+    ///
+    /// This check can be used before reading the file.
+    fn is_file_updated(&self) -> Result<bool>;
```
tareknaser, line 28:
```suggestion
    /// Check if storage is updated
    ///
    /// This check can be used before reading the file.
    fn is_storage_updated(&self) -> Result<bool>;
```

fs-storage/src/file_storage.rs
```diff
@@ -8,51 +9,110 @@ use std::{
...
 
 const STORAGE_VERSION: i32 = 2;
 const STORAGE_VERSION_PREFIX: &str = "version ";
 
-pub struct FileStorage {
+pub struct FileStorage<K, V> {
     label: String,
     path: PathBuf,
     timestamp: SystemTime,
+    pub value_by_id: BTreeMap<K, V>,
```
tareknaser, line 22: Why is `value_by_id` public? 
The API should only expose methods defined in the `BaseStorage` trait. 
For example, we have `as_ref` method that provides access to this field. 

Making this field public would enable users to bypass the methods and access it directly. If users modify this field directly, the `timestamp` wouldn't be updated, potentially leading to corrupt information on our side.

tareknaser, line 22: Nitpick
it could also be renamed to `data` or `entries`

fs-storage/src/file_storage.rs
```diff
@@ -8,51 +9,110 @@ use std::{
...
+{
+    fn get(&self, id: &K) -> Option<V> {
+        self.value_by_id.get(id).cloned()
+    }
+
+    fn set(&mut self, id: K, value: V) -> Result<()> {
+        self.value_by_id.insert(id, value);
+        self.timestamp = std::time::SystemTime::now();
+        Ok(())
+    }
```
tareknaser, line 92: This never returns an error, so it should not be `Result<()>`

fs-storage/src/file_storage.rs
```diff
@@ -8,51 +9,110 @@ use std::{
...
+
+    fn remove(&mut self, id: &K) -> Result<()> {
+        self.value_by_id.remove(id);
```
tareknaser, line 95:
```suggestion
        self.value_by_id.remove(id).ok_or_else(|| {
            ArklibError::Storage(self.label.clone(), "Key not found".to_owned())
        })?;
```

fs-storage/src/file_storage.rs
tareknaser, line 92: Do you think we should also write to disk here?
        Since we are updating the map, we should also update the information on file system

fs-storage/src/file_storage.rs
```diff
@@ -8,51 +9,110 @@ use std::{
...
+        self.timestamp = std::time::SystemTime::now();
+        Ok(())
+    }
```
tareknaser, line 98: Same here. 
Should we also write to disk?

fs-storage/src/base_storage.rs
```diff
@@ -0,0 +1,41 @@
...
+where
+    K: FromStr + Hash + Eq + Ord + Debug + Clone,
+    V: Debug + Clone,
+{
+    fn get(&self, id: &K) -> Option<V>;
+    fn set(&mut self, id: K, value: V) -> Result<()>;
+    fn remove(&mut self, id: &K) -> Result<()>;
+
+    /// Remove file at stored path
+    fn erase(&mut self) -> Result<()>;
```
tareknaser, line 18: I don't think we need mutable reference here 
```suggestion
    fn erase(&self) -> Result<()>;
```

fs-storage/src/base_storage.rs
```diff
@@ -0,0 +1,41 @@
...
+use std::str::FromStr;
+
+use data_error::Result;
+
+pub trait BaseStorage<K, V>
+where
+    K: FromStr + Hash + Eq + Ord + Debug + Clone,
+    V: Debug + Clone,
+{
+    fn get(&self, id: &K) -> Option<V>;
```
tareknaser, line 13:
```suggestion
    fn get(&self, id: &K) -> Option<&V>;
```

fs-storage/src/file_storage.rs
```diff
@@ -8,51 +9,110 @@ use std::{
...
+        + Ord
+        + Debug
+        + Clone
+        + serde::Serialize
+        + serde::de::DeserializeOwned,
+    V: Debug + Clone + serde::Serialize + serde::de::DeserializeOwned,
+{
+    fn get(&self, id: &K) -> Option<V> {
+        self.value_by_id.get(id).cloned()
+    }
```
tareknaser, line 86:
```suggestion
    fn get(&self, id: &K) -> Option<&V> {
        self.value_by_id.get(id)
    }
```

tareknaser (commented):
Thanks for putting this together

tareknaser:
I noticed that the current output from `FileStorage` isn't output a valid JSON file. The output is:
```
version 2
{"a":"1","b":"2","c":"3"}
```
It's also worth noting that the file doesn't have `.json` extension.
It probably should be 
```json
{
  "version": 2,
  "data": {
    "a": "1",
    "b": "2",
    "c": "3"
  }
}
```
This is not relevant to changes in this PR but we should fix it later

fs-storage/src/file_storage.rs
kirillt: Good question. We probably should write to disk periodically, not immediately. This is discussable.

Basically, updates could come often enough to produce several writes simultaneously. While we should be protected from dirty writes thanks to [fs-atomic-versions](https://github.com/ARK-Builders/ark-rust/tree/main/fs-atomic-versions), we still want to avoid unnecessary parallelism and batch updates together and write them once in a while, each 1 second.

On Kotlin side, this was achieved using [debounce](https://github.com/ARK-Builders/arklib-android/blob/377734870453889f3e9a8821bb1cac3f7f299ab8/lib/src/main/java/dev/arkbuilders/arklib/data/storage/BaseStorage.kt#L31)

kirillt: cc @twitu

kirillt:
Good note about JSON, the version header certainly came from our legacy format. We should convert it into JSON field.

And let's bump the version number to `3`.

fs-storage/src/base_storage.rs
Pushkarm029: cc @twitu

fs-storage/src/file_storage.rs
twitu: The proper way to do this would be to create an asynchronous job that periodically writes to disk. This is exactly what is done using debounce in Kotlin. However, adding async support will be outside the scope of this PR. Thankfully [async traits](https://blog.rust-lang.org/inside-rust/2023/05/03/stabilizing-async-fn-in-trait.html) have recently been stabilized in Rust. So it can be added in separately.

For now the safe but less efficient method will be to flush to disk on each write. On the other hand we can go for a slightly less safe but more efficient method by using a BufferedWriter to write data to the file. The BufferedWriter will automatically batch writes and flush to disk.

I propose going ahead with BufferedWriter approach.

fs-storage/src/file_storage.rs
```diff
@@ -8,51 +9,111 @@ use std::{
...
+    K: serde::Serialize,
+    V: serde::Serialize,
+{
     /// Create a new file storage with a diagnostic label and file path
     pub fn new(label: String, path: &Path) -> Self {
         Self {
             label,
             path: PathBuf::from(path),
             timestamp: SystemTime::now(),
+            data: BTreeMap::new(),
```
twitu, line 36: Data should be loaded into the tree from the given path when the FileStorage is created here.

fs-storage/src/base_storage.rs
twitu: By exposing it directly we can provide the user all the immutable operations for free. Otherwise if a user wants to use the btree as an iterator we will have to add in a wrapper. Similarly for each extra function that needs to be supported `ark-rust` will have to add a wrapper.

In any case, I was going for something along these lines.

```
pub trait BaseStorage<K, V>: AsRef<BTreeMap<K, V>>
where
  ...
```

This way anything that implements BaseStorage must also be referencable as a BtreeMap which fits are notion that BaseStorage is a virtualization of a key value mapping stored on disk.

[trait inheritance](https://users.rust-lang.org/t/extending-traits/1802)
[super trait](https://doc.rust-lang.org/book/ch19-03-advanced-traits.html#using-supertraits-to-require-one-traits-functionality-within-another-trait)

tareknaser: Got it. Thanks for the clarification.

fs-storage/src/base_storage.rs
```diff
@@ -0,0 +1,36 @@
...
+use std::str::FromStr;
+
+use data_error::Result;
+
+pub trait BaseStorage<K, V>: AsRef<BTreeMap<K, V>>
+where
+    K: FromStr + Hash + Eq + Ord + Debug + Clone,
+    V: Debug + Clone,
+{
+    fn get(&self, id: &K) -> Option<&V>;
```
twitu, line 13: `get` will automatically be provided by `AsRef<Btree..`. An explicit `get` is not necessary.

fs-storage/src/file_storage.rs
```diff
@@ -120,47 +188,26 @@ impl FileStorage {
...
-                    "Storage version mismatch: expected {}, found {}",
-                    STORAGE_VERSION, version
-                ),
-            ));
-        }
-
-        Ok(())
+impl<K, V> AsRef<BTreeMap<K, V>> for FileStorage<K, V>
+where
+    K: FromStr
```
twitu, line 199: Do you need all these traits for this implementation? It's best to define only the trait constraints that are necessary.

twitu (approved):
Minor changes. Rest looks good :100:

This will make it easy to add in the conflict resolution too!

fs-storage/src/base_storage.rs
Pushkarm029: 2 min

fs-storage/src/file_storage.rs
Pushkarm029: done

twitu (approved):

tareknaser (commented):
Thank you for putting this together. I think we are ready to merge. 
I noticed some functions are not tested.

fs-storage/src/base_storage.rs
tareknaser: `get` is indeed not necessary but I find it convenient to have. Additionally, if a struct implements `get`, it implies that it also implements `set`

instead of doing
```rust
let data = file_storage.as_ref();
let key_1_value = data.get("key1").unwrap();
```
we can just do
```rust
let key_1_value = file_storage.get("key1").unwrap();
```
What do you think?

fs-storage/src/file_storage.rs
```diff
@@ -1,58 +1,115 @@
...
+        self.timestamp = std::time::SystemTime::now();
+        self.write_fs()
+            .expect("Failed to write data to disk");
     }
 
-    /// Check if underlying file has been updated
-    ///
-    /// This check can be used before reading the file.
-    pub fn is_file_updated(&self) -> Result<bool> {
+    fn remove(&mut self, id: &K) -> Result<()> {
```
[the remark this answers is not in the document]
Pushkarm029: Added Now, please have a look

fs-storage/src/file_storage.rs
```diff
@@ -219,4 +226,28 @@ mod tests {
         }
         assert_eq!(storage_path.exists(), false);
     }
+
+    #[test]
+    fn test_file_storage_is_storage_updated() {
```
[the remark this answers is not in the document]
tareknaser: Tracing the error back. seems like the it is because the nanoseconds don't match. `SystemTime` stores the time in nanosecond precision
```sh
running 1 test
Self timestamp: SystemTime { tv_sec: 1713553086, tv_nsec: 320082935 }
File timestamp: SystemTime { tv_sec: 1713553086, tv_nsec: 320194727 }
thread 'file_storage::tests::test_file_storage_is_storage_updated' panicked at fs-storage/src/file_storage.rs:243:9:
  left: true
 right: false
```

I don't think we need nanosecond precision. Maybe we can just compare seconds
```rust
fn is_storage_updated(&self) -> Result<bool> {
	let file_timestamp = fs::metadata(&self.path)?.modified()?;
	let file_time_secs = file_timestamp
		.duration_since(SystemTime::UNIX_EPOCH)
		.unwrap() // SAFETY: We can safely unwrap because the timestamp is always after the epoch
		.as_secs();
	let self_time_secs = self
		.timestamp
		.duration_since(SystemTime::UNIX_EPOCH)
		.unwrap()
		.as_secs();
	Ok(file_time_secs > self_time_secs)
}
```
But you might have a better idea.

fs-storage/src/base_storage.rs
```diff
@@ -0,0 +1,27 @@
+use data_error::Result;
+use std::collections::BTreeMap;
+
+pub trait BaseStorage<K, V>: AsRef<BTreeMap<K, V>> {
+    fn set(&mut self, id: K, value: V);
```
tareknaser, line 5: I think we should have `get` too here
See https://github.com/ARK-Builders/ark-rust/pull/23#discussion_r1570703803

fs-storage/src/base_storage.rs
twitu: A get generally implies a set, especially coming from the OOP paradigm. In this case, we are getting a get it is from the BtreeMap which the FileStorage can be treated as. So I don't see a need for an explicit get especially if the get is calling the BtreeMap get again.

fs-storage/src/file_storage.rs
```diff
@@ -1,58 +1,115 @@
...
             path: PathBuf::from(path),
             timestamp: SystemTime::now(),
+            data: BTreeMap::new(),
+        };
+
+        // Load the data from the file
+        file_storage.data = match file_storage.read_fs() {
+            Ok(data) => data,
+            Err(err) => {
+                log::error!("Error reading storage file: {}", err);
```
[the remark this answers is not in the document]
twitu: Indeed, the first time a FileStorage is created the file doesn't exist. Perhaps that can be checked here. If the file doesn't exist when creating the FileStorage it's being created for the first time. So it can start with an empty BTreeMap.

fs-storage/src/base_storage.rs
Pushkarm029: https://github.com/ARK-Builders/ark-rust/pull/23#discussion_r1573135363

fs-storage/src/base_storage.rs
tareknaser: No problem, it's a detail. 
Let's focus on what matters most. Moving forward with the PR

tareknaser (approved):
Thanks for putting this together

fs-storage/src/file_storage.rs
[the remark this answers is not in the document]
kirillt: @tareknaser probably we need more granular timestamps. It's ok to merge it like this, but let's think about such scenario:
1. We scan folder at the moment `HH:MM:ss:xxx`, where `xxx` is milliseconds part.
2. A file `cat.jpg` is modified at the moment `HH:MM:ss:yyy`.

If we truncate `xxx` and `yyy` then on next scan we'll skip `cat.jpg` and we miss the update.

As a workaround, it seems possible rounding `HH:MM:ss:xxx` always to the lowest side, and `HH:MM:ss:yyy` always to the highest side. This would result in repeated updates for the files which were updated in the same second when scan happened, but it's better than to lose an update.

Also, I think we have discussed somewhere the smallest division of time that makes sense for file systems... Nanoseconds are too low, but what about milliseconds?

@twitu

PR #26, opened 2024-04-09
`fs-storage` implemented trait for conflict resolution

### LINKED ISSUE #19: `fs-storage`: Port Monoid-based conflicts resolution from Kotlin
Port [Monoid structure](https://github.com/ARK-Builders/arklib-android/blob/main/lib/src/main/java/dev/arkbuilders/arklib/data/storage/Monoid.kt) used during conflicts resolution. It's naive approach but it's simpler than CRDT.

Pushkarm029:
resolves #19


fs-storage/src/crdt.rs
```diff
@@ -0,0 +1,20 @@
+// Currently, we have three structures: Tags (HashSet), Properties (HashSet), Score (int).
+// In fact, HashSet already implements a union function,
+// so only a special function for integers is needed.
+// CRDTs can be considered later when we need to add structures that require
+// more powerful combine semantics.
+
+pub trait CRDT<V> {
```
kirillt, line 7: So far it looks only like a Monoid :)

Pushkarm029: Oh, I forgot to rename it.

twitu (changes requested):
You also want to implement the Monoid trait for some common values. For now integers and HashSets are good since those are being commonly used in the Kotlin implementation as discussed in #19 

Also you'll want to merge #23 in first so that we have the BaseStorage merged. Then add the monoid based conflict resolution logic to BaseStorage in this PR.

fs-storage/src/file_storage.rs
```diff
@@ -120,47 +182,32 @@ impl FileStorage {
...
-
-        let version = header[STORAGE_VERSION_PREFIX.len()..]
-            .parse::<i32>()
-            .map_err(|_err| {
-                ArklibError::Storage(
-                    self.label.clone(),
-                    "Failed to parse storage version".to_owned(),
-                )
-            })?;
+impl<K, V> Monoid<V> for FileStorage<K, V>
```
tareknaser, line 197: The concept of "combining" two `FileStorage` structures needs a clear definition. Does it involve merging key-value pairs? How are conflicts handled? The current `combine` function only compares single values.

---
To get a sense of what I mean:

In a `Monoid` trait for integers, addition combines elements, with 0 as the neutral element. For example, combining `3` and `5` is `8`. This operation satisfies:
- Associativity: Grouping elements doesn't change the result (e.g., `(3 + 5) + 2` = `3 + (5 + 2)`).
- Neutral element: Adding 0 to any element leaves it unchanged (e.g., `3 + 0` = 3).

The `FileStorage<K, V>` struct holds a collection of key-value pairs. The `Monoid` trait doesn't naturally apply to this structure because:
What does combining two `FileStorage` instances even mean? Does it involve:
- Merging all key-value pairs?
- Overwriting existing keys with values from the second storage?
- How does it handle duplicate keys?

tareknaser (commented):
Thanks
Could you rebase on top of main?

fs-storage/src/file_storage.rs
```diff
@@ -169,7 +216,7 @@ mod tests {
     use std::collections::BTreeMap;
     use tempdir::TempDir;
 
-    use crate::file_storage::FileStorage;
+    use crate::{base_storage::BaseStorage, file_storage::FileStorage};
 
     #[test]
```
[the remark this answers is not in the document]
Pushkarm029: i will add it after finalizing the main code.

fs-storage/src/monoid.rs
```diff
@@ -0,0 +1,17 @@
+// Currently, we have three structures: Tags (HashSet), Properties (HashSet), Score (int).
+// In fact, HashSet already implements a union function,
+// so only a special function for integers is needed.
+// CRDTs can be considered later when we need to add structures that require
+// more powerful combine semantics.
+
+pub trait Monoid<V> {
+    fn neutral() -> V;
```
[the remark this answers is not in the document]
Pushkarm029: same doubt
cc @twitu

fs-storage/src/file_storage.rs
Pushkarm029: We are implementing monoids for Tags, Properties, and Scores. While Tags and Properties are of type `HashSet`, they have already implements a union function. So, only the score left of type `INT` requires the particular function to be managed. That's why, instead of adding, we are comparing values(same as in Kotlin) - More info for scores [here](https://github.com/ARK-Builders/arklib-android/blob/main/lib/src/main/java/dev/arkbuilders/arklib/user/score/Score.kt#L9).

ref: https://github.com/ARK-Builders/ark-rust/issues/19#issuecomment-2018964840

tareknaser: Thanks for the explanation, Pushkar.
I am still trying to wrap my head around the end goal.

I think adding a test would help clarify. 
Could you include a test where you create two `FileStorage` instances, each with its own tags, properties, and scores, and then use `FileStorage::combine_all()` to merge them?

fs-storage/src/monoid.rs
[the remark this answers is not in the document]
twitu: There's no practical difference as such. It's just that Monoid is defined as having a combine and an empty operator. This is consistent across all language implementations (for e.g. [haskell](https://hackage.haskell.org/package/base-4.19.1.0/docs/Data-Monoid.html)). It's better to be consistent with the definition of a monoid. Otherwise, it might be better to call it something else.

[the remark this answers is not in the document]
kirillt: Yep, `neutral` is from standard Monoid definitions. It's coming from algebra, where it is defined over some operation `+` together with some set `X`. Pair (`+`, `X`) must possess 2 properties to be considered a Monoid:
1. `X` contains a `neutral` element, i.e. such `0` that `0 + a = a + 0 = 0` for any `a`.
2. `+` is associative, i.e. `a + (b + c)` is same as `(a + b) + c`.

It's the simplest and the easiest to use algebra abstraction :) It helps to define something what can be summed. E.g. strings, vectors, sets, etc. Defining it as a trait helps to build merge pattern, abstract from particular data.

fs-storage/src/file_storage.rs
kirillt: Yeah, that's for conflicts resolution. I'm not sure yet if `FileStorage` itself should be a `Monoid` (need to review the code for this).

What was done before in Kotlin is:
1. Take entries with unique keys from both maps, dump them into the resulting map.
2. For clashing entries, i.e. attached to same keys, we `combine` their values.

I think we can declare `FileStorage` as `Monoid`, if it's convenient. But it isn't necessary.

Pushkarm029:
sorry for delay

fs-storage/src/file_storage.rs
```diff
@@ -105,26 +105,36 @@ where
...
+                let self_modified = self
+                    .modified
+                    .duration_since(SystemTime::UNIX_EPOCH)
+                    .unwrap()
+                    .as_secs();
+
+                Ok(fs_modified != self_modified)
+            },
+            Err(e) => {
+                //todo: log the error
```
kirillt, line 134: @Pushkarm029 could you double-check my refactoring of this function, optimize if possible and add error logs? This function was not really correct before, and to make it completely correct we need [atomic versions](https://github.com/ARK-Builders/ark-rust/issues/3) but at least we should flag correctly when the data doesn't match the disk. This can be either because of updates to the in-memory map, or because of updates to the filesystem.

fs-storage/src/file_storage.rs
```diff
@@ -280,19 +331,58 @@ mod tests {
 
         let mut file_storage =
             FileStorage::new("TestStorage".to_string(), &storage_path);
+        file_storage.write_fs().unwrap();
+        assert_eq!(file_storage.is_outdated().unwrap(), false);
 
         file_storage.set("key1".to_string(), "value1".to_string());
-        assert_eq!(file_storage.is_storage_updated().unwrap(), false);
+        //todo: we need to add 1ms delays to make the test pass
```
kirillt, line 338: @Pushkarm029 could you also complete this test case? The `is_updated` functions wasn't really tested properly.

fs-storage/src/file_storage.rs
```diff
@@ -224,13 +234,54 @@ where
     }
 }
 
+/// Combine two `FileStorage` instances into a new instance.
+pub fn combine_file_storages<K, V>(
+    a: &FileStorage<K, V>,
+    b: &FileStorage<K, V>,
+) -> FileStorage<K, V>
```
kirillt, line 241: Let's move this function to `BaseStorage` trait:
```
    fn merge_from(&mut self, other: AsRef<BTreeMap<K, V>>) -> Result<()>;
```

This makes it explicit what path is used to store the result. Also, making it a function in the trait allows very flexible usage. For isntance, we could update a file storage using values from a folder storage (or future chunked storage).

fs-storage/src/file_storage.rs
```diff
@@ -224,13 +234,54 @@ where
...
+
+    for (key, value) in &a.data.entries {
+        combined_entries.insert(key.clone(), value.clone());
+    }
+
+    for (key, value) in &b.data.entries {
+        if let Some(existing_value) = combined_entries.get(key) {
+            let resolved_value = if value > existing_value {
+                value.clone()
+            } else {
```
kirillt, line 256: Here, we need to use `combine` function from Monoid, whenever `value` isn't equal to `existing_value`.

kirillt (changes requested):
Good start, but Monoid isn't really used during combining storages. Also, let's generalize the storage combining function and move it to `BaseStorage` trait. Sorry for adding couple of `todo`s in the code, they are not related to Monoids issue, but I noticed it during the code review.

Pushkarm029:
@kirillt 💖Thanks for the review. 🚀working on it

fs-storage/src/file_storage.rs
twitu: Using metadata to check outdatedness is a valid solution. In fact, mature tools like rsync do the same. Perhaps we can increase the time resolution to use milliseconds with `as_millis`.

Also about atomic versions, I think it brings in significant complexity to the implementation so it can be put behind a feature flag. More, many use cases that involve user interaction it is very unlikely for their to be race conditions related to timestamps. Such use cases can use this (default) implementation. For automated updated use cases, the feature flag can be turned on and use atomic versions.

fs-storage/src/file_storage.rs
```diff
@@ -224,13 +234,54 @@ where
...
+        }
+    }
+
+    FileStorage {
+        label: a.label.clone(),
+        path: a.path.clone(),
+        modified: SystemTime::now(),
+        data: FileStorageData {
+            version: STORAGE_VERSION,
+            entries: combined_entries,
```
twitu, line 271: The `combined_entries` also need to be written to disk, otherwise it will be out of sync with data on disk.

fs-storage/src/file_storage.rs
Pushkarm029: Even after adding delay, it will not pass because `filestorage.set` internally calls `write_fs`, so it no longer remains outdated. I will try to come up with another approach.

Pushkarm029:
I made some changes related to `monoids`, please review.

fs-storage/src/file_storage.rs
```diff
@@ -234,6 +271,23 @@ where
     }
 }
 
+impl<K, V> Monoid<V> for FileStorage<K, V>
```
twitu, line 274: As you can see, this Monoid implementation does not use any property or value of the FileStorage. It doesn't even use the key. It only relies on the properties of the `V` value. In such a case, it's probably better to remove the Monoid implementation for FileStorage.

Instead, add a trait condition for V in `merge_from` like

```rust
    fn merge_from(&mut self, other: impl AsRef<BTreeMap<K, V>>) -> Result<()>
        where V: Monoid 
        {
```

And the the `resolved_value` can be decided based on the Monoid implementation of `V`. In fact, this approach is closer to your idea of using Monoid as a helper function for `FileStorage`.

Pushkarm029: Done;
Currently, I only implemented monoid for integers and strings. Let me know if we need more.

fs-storage/src/file_storage.rs
twitu: Perhaps I was too eager in making `set` immediately sync the file system. It'll be more useful to allow the library user to decide whether they want to write to disk immediately after `set`ting a value or use a different strategy like batching writes.

Once `write_fs` is moved out of `set` this test will also be valid.

fs-storage/src/monoid.rs
```diff
@@ -0,0 +1,46 @@
...
+            *a
+        } else {
+            *b
+        }
+    }
+}
+
+impl Monoid<String> for String {
+    fn neutral() -> String {
+        "".to_string()
```
twitu, line 40: Use `String::new()` here.

fs-storage/src/monoid.rs
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    fn combine(a: &String, b: &String) -> String {
```
twitu, line 43: Also this way of combining will be incredibly inefficient for Strings because it'll reallocate on each combine. Typically practical monoid implementations also include a `combine_all` that takes a an iterator of values and combines them in an efficient manner. Like Python's join operator that takes an iterator of strings instead of `+` each string individually.

This is not relevant immediately but something to keep in mind when monoids are used to combine over iterators (aka folds).

twitu (commented):
Looking good!

fs-storage/src/file_storage.rs
kirillt: Yes, atomic versions is more complicated mechanism. We don't need to implement it in this PR, but it will be a good thing.

Rsync is a bit different use-case if look at it on single-machine scale, because it only reads state of the filesystem (write is happening on a remote machine). And we do both reading and writing within single machine. But I think timestamp solution is good enough for now, the only problem is that we only can detect when we need an update, but we can't figure out which side (fs or ram) is behind the other. Because changes could be made on both sides.

But atomic versions probably don't solve this case, we would still have diverged versions and sync them using monoids 🤔

fs-storage/src/file_storage.rs
kirillt: I didn't know this when I refactored the test case, but good that we stumbled upon this. Yes, better to have flexibility regarding when changes are pushed to the disk. In the prototype (arklib-android and Navigator app), we used debounce to write the changes each 1 second or so, because each write was performed within a coroutine and user can create several or more actions per seconds. So, coroutines were racing to write different data into same file, resulting in half-written file... effectively erasing half of the file.

Actually, that was the moment when we realized the need for some protection. With "write into temp file, then move" method, we still can lose some new data (but not everything), with simple mutex we still can overwrite newer data with the older.

Atomic versions came up at this moment, because with them we don't write older versions over the newer versions (mutex + counter + new location every time).

fs-storage/src/file_storage.rs
```diff
@@ -90,41 +85,53 @@ where
...
-        Ok(file_time_secs > self_time_secs)
+    /// to determine if the storage file is up-to-date.
+    fn is_outdated(&self) -> Result<bool> {
+        match fs::metadata(&self.path) {
+            Ok(metadata) => {
+                let fs_modified = metadata
+                    .modified()?
+                    .duration_since(SystemTime::UNIX_EPOCH)
+                    .unwrap()
+                    .as_secs();
```
kirillt, line 117: Probably a convenience function would be nice to have, to reduce code duplication now and in future.

fs-storage/src/base_storage.rs
```diff
@@ -24,4 +24,8 @@ pub trait BaseStorage<K, V>: AsRef<BTreeMap<K, V>> {
     /// Remove all persisted data
     /// by pre-configured location in the file-system.
     fn erase(&self) -> Result<()>;
+
+    /// Merge two storages instances
+    /// and write the result to the filesystem.
```
kirillt, line 29: Here should be same as in other storage mutations, i.e. we should enable user to control when writes happen.

@twitu should we merge into in-memory model, but not write until the user calls `fn write_fs`?
If we have merging in `fn write_fs` then we'd do merging twice...

fs-storage/src/file_storage.rs
```diff
@@ -213,6 +220,30 @@ where
...
+        for (key, value) in other_entries {
+            if let Some(existing_value) = self.data.entries.get(key) {
+                let resolved_value = V::combine(existing_value, value);
+                self.data
+                    .entries
+                    .insert(key.clone(), resolved_value);
+            } else {
+                self.data
+                    .entries
+                    .insert(key.clone(), value.clone());
```
kirillt, line 239: Sorry for the nitpick, but here is some small duplication

fs-storage/src/base_storage.rs
kirillt: Or should we just allow the exception for `merge_from` for this moment, and call `write_fs` from it?

twitu: Now that we've removed `write_fs` from `set`, it'll be consistent to remove it from `merge_from` as well. This way we can leave the writes entirely up to the application consuming ark-rust logic.

fs-storage/src/base_storage.rs
```diff
@@ -11,7 +11,13 @@ pub trait BaseStorage<K, V>: AsRef<BTreeMap<K, V>> {
...
     /// i.e. that the internal mapping is consistent
     /// with the data in the filesystem.
-    fn is_storage_updated(&self) -> Result<bool>;
+    /// Returns:
+    /// - `Ok(true)` if the on-disk data is outdated compared to the in-memory data.
+    /// - `Ok(false)` if the on-disk data is up-to-date with the in-memory data.
+    /// If there is an error retrieving the file metadata (e.g., the file doesn't
+    /// exist), `Ok(true)` is returned, indicating that the on-disk data is outdated
+    /// or non-existent.
+    fn is_outdated(&self) -> Result<bool>;
```
tareknaser, line 20: Returning `Ok(true)` if there is an error retrieving the file metadata might hide the fact that an error occurred. 

Would it be clearer to separate the cases where the data is outdated from when the file doesn't exist in the first place? We could return `Err(_)` if the file doesn't exist.

twitu (approved):

kirillt:
Last change:
1. Returning `Err` in case of any error, now we don't separate case of non-existing file or other error. `Ok(true)` means that the storage exists and it's outdated (or our model is outdated). Developers using the crate are free to try syncing if they encounter `Err`.
2. Renamed the `is_outdated` function and enhanced comments to highlight the symmetry between the model and storage. Any of them could be outdated, or both of them. New name is `needs_syncing`.

kirillt (approved):

kirillt:
Thanks @Pushkarm029 and @twitu as well as @tareknaser! Great team work 👍

Issue #29, opened 2024-04-18
Generic Resource Identifiers with `ResourceIdTrait`


tareknaser:
## Description
Merge changes from [https://github.com/ARK-Builders/arklib/pull/90](https://github.com/ARK-Builders/arklib/pull/90) into `ark-rust`, with the cryptographic and non-cryptographic methods of operation pushed behind feature flags in relevant crates.
## Motivation
Cryptographic hash functions ensure unique IDs for each resource, so there will be no need for collision handling in `fs-index`. To support non-cryptographic hash functions as well, we're proposing placing them behind feature flags.
### Changes to `ResourceIdTrait` from [https://github.com/ARK-Builders/arklib/pull/90](https://github.com/ARK-Builders/arklib/pull/90)
- Simplified to include only methods to get `ResourceId` from path or data bytes.
```rust
pub trait ResourceIdTrait {
    /// Associated type representing the hash used by this resource identifier.
    type HashType: Debug; // + Other traits we want to enforce

    /// Computes the resource identifier from the given file path
    fn from_path<P: AsRef<Path>>(file_path: P) -> Result<Self::HashType>;

    /// Computes the resource identifier from the given bytes
    fn from_bytes(data: &[u8]) -> Result<Self::HashType>;
}
```
- `ResourceId` doesn't need a field `data-size`
### The 2 features that should be available:
- `cryptographic-hash`: Uses cryptographic hash functions (e.g., `blake3`) to define `ResourceId`, enabled by default.
- `non-cryptographic-hash`: Uses non-cryptographic hash functions (e.g., `crc32fast`) for `ResourceId`.
### Commands to build 
- Building for cryptographic hash: `cargo build --release`
- Building for non-cryptographic hash: `cargo build --release --workspace --features non-cryptographic-hash --no-default-features`
### Relevant crates:
- `data-resource`: Main crate exporting different `ResourceId` types based on enabled features. Implementation should cleanly export `ResourceId` with enforced `HashType` traits, avoiding impact on `fs-index`.
- `fs-index`: Handles collisions and index management if non-cryptographic hash `ResourceId` is exported by `data-resource`.
- `ark-cli`: Includes `ark collisions` command and other relevant commands if non-cryptographic hash `ResourceId` is used in `data-resource`.  
## Follow up work:
- Update the CI to include sanity checks for build, clippy, formatting, and testing in both cryptographic and non-cryptographic hash modes of operation.
## Side note:
[...]

PR #32, opened 2024-04-21
Generic Resource Identifiers with `ResourceIdTrait`

tareknaser:
## Description
This PR ports changes from https://github.com/ARK-Builders/arklib/pull/90 into `ark-rust`.
It represents the first milestone of #29
## Changes
- Introduce `ResourceIdTrait`:
	- Hash type restrictions
    - Calculate hash values from path or bytes
- Add compilation features for `data-resource`:
    - "non-cryptographic-hash" and "cryptographic-hash" (default) features
- Implement `blake3::ResourceId` and `crc32::ResourceId`:
    - Include `blake3::ResourceId` under the "cryptographic-hash" feature
    - Include `crc32::ResourceId` under the "non-cryptographic-hash" feature
- Define benchmarks for hash creation:
    - Create benchmarks for hash generation from paths and bytes
- Update other crates in `ark-rust` workspace:
    - Transition them to use the "non-cryptographic-hash" feature for now, because previously they were using crc32 hash values directly


data-resource/Cargo.toml
```diff
@@ -8,23 +8,35 @@ name = "data_resource"
...
-name = "compute_bytes_benchmark"
-harness = false
-path = "benches/compute_bytes_benchmark.rs"
+
+# Features:
+# - `non-cryptographic-hash`: Uses a non-cryptographic hash functions to define `ResourceId`
+# - `cryptographic-hash`: Uses a cryptographic hash functions to define `ResourceId`
+[features]
+non-cryptographic-hash = ["crc32fast"]
+cryptographic-hash = ["blake3", "hex"]
```
kirillt, line 40: What does `"hex"` mean here?

data-resource/src/blake3.rs
```diff
@@ -0,0 +1,84 @@
...
+    PartialEq,
+    PartialOrd,
+    Hash,
+    Clone,
+    Copy,
+    Debug,
+    Deserialize,
+    Serialize,
+)]
+pub struct ResourceId {}
```
kirillt, line 28: Is it stub for future implementation? We have `"cryptographic-hash"` enabled by default, so we should provide the default implementation.

data-resource/src/crc32.rs
```diff
@@ -0,0 +1,75 @@
...
+    PartialEq,
+    PartialOrd,
+    Hash,
+    Clone,
+    Copy,
+    Debug,
+    Deserialize,
+    Serialize,
+)]
+pub struct ResourceId {}
```
kirillt, line 27: It was like this:
```
pub struct ResourceId {
    pub data_size: u64,
    pub hash: u32,
}
```

By the way, could it be useful to have definition like this shared across implementations?
```
pub struct ResourceId<H> {
    pub data_size: u64,
    pub hash: H,
}
```

Then we would have `ResourceId` type everywhere in the client code without the need for aliases.

ark-cli/src/main.rs
```diff
@@ -5,6 +5,9 @@ use std::str::FromStr;
 
 use data_pdf::{render_preview_page, PDFQuality};
 use data_resource::ResourceId;
+use data_resource::ResourceIdTrait;
+pub(crate) type HashType = <ResourceId as ResourceIdTrait>::HashType;
```
kirillt, line 9: We should provide more descriptive name to the end-user. Ideally, `ResourceId`.

By the way, if we make `HashType` a parameter of common `ResourceId` structure, we could avoid fixing the name for it (see the comment about generic `ResourceId` structure).

data-resource/src/blake3.rs
```diff
@@ -0,0 +1,84 @@
...
+
+impl ResourceIdTrait for ResourceId {
+    type HashType = String;
```
kirillt, line 31: Probably `Vec<u64>` would be better.

data-resource/Cargo.toml
tareknaser: It's the [hex](https://crates.io/crates/hex) crate
We currently use it in `cryptographic-hash` feature because we are hex encoding the `ResourceId` and storing it as a string

data-resource/src/blake3.rs
tareknaser: We can't directly use `Vec<u8>` since it doesn't come with `Display` or `FromStr` implementations, which we typically expect `ResourceId` to have in other crates. To handle this, we can wrap it in a struct
```rust
#[derive(
    Hash, Ord, PartialOrd, Eq, PartialEq, Clone, Debug, Serialize, Deserialize,
)]
pub struct Hash(Vec<u8>);

impl Display for Hash {
}

impl FromStr for Hash {
}
```
I gave it a try in this commit https://github.com/ARK-Builders/ark-rust/commit/382aa9bb51e8c0b4622a5a08988a31c10496796d

---
Ran benchmarks comparing it to using a plain `String` for the hash (hex encoded), and here are the results:
```sh
Gnuplot not found, using plotters backend
resource_id_creation/compute_from_path:../test-assets/lena.jpg
                        time:   [194.53 µs 195.35 µs 196.66 µs]
                        change: [+0.0968% +0.4659% +1.0237%] (p = 0.02 < 0.05)
                        Change within noise threshold.
Found 5 outliers among 100 measurements (5.00%)
  1 (1.00%) low mild
  1 (1.00%) high mild
  3 (3.00%) high severe
resource_id_creation/compute_from_path:../test-assets/test.pdf
                        time:   [1.6235 ms 1.6256 ms 1.6287 ms]
                        change: [+0.2701% +0.4232% +0.5821%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 6 outliers among 100 measurements (6.00%)
  3 (3.00%) high mild
  3 (3.00%) high severe
resource_id_creation/compute_from_bytes:small
                        time:   [1.2565 µs 1.2570 µs 1.2576 µs]
                        change: [-9.3952% -9.2482% -9.1129%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 12 outliers among 100 measurements (12.00%)
  3 (3.00%) low severe
  5 (5.00%) low mild
  2 (2.00%) high mild
  2 (2.00%) high severe
resource_id_creation/compute_from_bytes:medium
                        time:   [37.043 µs 37.055 µs 37.067 µs]
                        change: [-0.7187% -0.4734% -0.2778%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 11 outliers among 100 measurements (11.00%)
  2 (2.00%) low severe
  1 (1.00%) low mild
  6 (6.00%) high mild
  2 (2.00%) high severe
resource_id_creation/compute_from_bytes:large
                        time:   [592.64 µs 592.84 µs 593.05 µs]
                        change: [-2.1562% -1.2193% -0.5402%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 6 outliers among 100 measurements (6.00%)
  1 (1.00%) low severe
  1 (1.00%) low mild
  3 (3.00%) high mild
  1 (1.00%) high severe
```
The performance seems relatively unchanged.

  
If you still prefer storing the hash in `Vec<u8>` for other reasons, I can directly cherrypick the commit to this PR.

data-resource/src/blake3.rs
tareknaser: The `ResourceId` struct doesn't hold any data because there is no need to. Instead, methods like `from_path` and `from_bytes` return hash values directly. 

The `ResourceIdTrait` simply enforces some restrictions on types implementing it, to allow us to use it from other crates without worrying about the underlying hash function.

data-resource/src/crc32.rs
tareknaser: We can use a generic type instead of an associated type
```rust
pub trait ResourceIdTrait<T>
where
    T: Debug
        + Display
        + FromStr
        + Clone
        + PartialEq
        + Eq
        + Ord
        + PartialOrd
        + Hash
        + Serialize,
{
    /// Computes the resource identifier from the given file path
    fn from_path<P: AsRef<Path>>(file_path: P) -> Result<T>;
    /// Computes the resource identifier from the given bytes
    fn from_bytes(data: &[u8]) -> Result<T>;
}
```
and in `data-resource/src/crc32.rs`, we can implement it 
```rust
impl ResourceIdTrait<u32> for ResourceId {
    fn from_path<P: AsRef<Path>>(file_path: P) -> Result<u32> {
```
I'm not entirely sure how that change would affect our situation.

Btw, do you think we should keep `data_size` as a field of `ResourceId`?

tareknaser: Another point to keep in mind:
With `pub trait ResourceIdTrait<T> where T: Debug`, we'd need to manually annotate types when importing `ResourceId` in other crates. 

This isn't ideal as we want to switch between "non-cryptographic-hash" and "cryptographic-hash" implementations without manual code changes if possible.

ark-cli/src/main.rs
tareknaser: > We should provide more descriptive name to the end-user. Ideally, `ResourceId`.

Great point. totally agree. 
Renamed `HashType` to `ResourceId`. This shifts the long use statement from other crates to `data-resource`

data-resource/Cargo.toml
tareknaser: I made some changes to the dependency structure to be more clear. Let me know your thoughts on it.

data-resource/src/blake3.rs
kirillt: @tareknaser benchmarking is great. If it doesn't reveal any performance benefit then String is fine.

Could you also benchmark with `Vec<u64>` instead of `Vec<u8>`? Only `compute_from_bytes ` benchmarks is enough.

tareknaser: This commit modifies the data type from `Vec<u8>` to `Vec<u64>`: https://github.com/tareknaser/ark-rust/commit/8f0b3473ae567b303e4b8b21edfba199054e8c28 
Here are the benchmarks for moving from hash of type `String` to `Vec<u64>`
```sh
resource_id_creation/compute_from_bytes:small
                        time:   [241.08 ns 241.85 ns 242.89 ns]
                        change: [+0.0702% +0.5050% +0.9694%] (p = 0.02 < 0.05)
                        Change within noise threshold.
Found 9 outliers among 100 measurements (9.00%)
  2 (2.00%) low mild
  3 (3.00%) high mild
  4 (4.00%) high severe
resource_id_creation/compute_from_bytes:medium
                        time:   [15.043 µs 15.088 µs 15.148 µs]
                        change: [-0.2250% +0.3452% +1.1816%] (p = 0.35 > 0.05)
                        No change in performance detected.
Found 15 outliers among 100 measurements (15.00%)
  3 (3.00%) low severe
  1 (1.00%) low mild
  1 (1.00%) high mild
  10 (10.00%) high severe
resource_id_creation/compute_from_bytes:large
                        time:   [246.52 µs 247.14 µs 248.10 µs]
                        change: [-0.4475% +0.0280% +0.4609%] (p = 0.91 > 0.05)
                        No change in performance detected.
Found 11 outliers among 100 measurements (11.00%)
  1 (1.00%) low severe
  1 (1.00%) low mild
  3 (3.00%) high mild
  6 (6.00%) high severe

     Running benches/index_build_benchmark.rs (target/release/deps/index_build_benchmark-6ce1452d366f8dbc)
Gnuplot not found, using plotters backend
index_build/index_build/../test-assets/
                        time:   [1.0621 ms 1.0648 ms 1.0686 ms]
                        change: [-0.3661% -0.0233% +0.3024%] (p = 0.89 > 0.05)
                        No change in performance detected.
Found 9 outliers among 100 measurements (9.00%)
  1 (1.00%) low severe
  2 (2.00%) high mild
  6 (6.00%) high severe
```
There is no change in performance. 

Do you think we should stick with `String` since it's simpler to read and maintain?

kirillt: @tareknaser could you perform last benchmarks with 10mb data size? But overall it seems that `Vec<u64>` is a premature optimization. Maybe it's different on really huge files like 1GB but we target on 512kb - 15mb range as the most common.

tareknaser: Sure. 
I ran some benchmarks on a 13 MB PDF file, using commit 8f0b347.
Here are the benchmarks for the transition from a hash of type `String` to `Vec<u64>`.

```sh
tareknasser@Tareks-MacBook ark-rust % ls -l /Users/tareknasser/Documents/books/Database_System_Concepts.pdf
-rw-r--r--@ 1 tareknasser  staff  13802981 Dec 28 20:15 /Users/tareknasser/Documents/books/Database_System_Concepts.pdf
tareknasser@Tareks-MacBook ark-rust % cargo bench 
   Compiling data-resource v0.1.0 (/Users/tareknasser/Documents/workspace/rust/arkbuilders/ark-rust/data-resource)
   Compiling fs-index v0.1.0 (/Users/tareknasser/Documents/workspace/rust/arkbuilders/ark-rust/fs-index)
    Finished bench [optimized] target(s) in 2.41s
     Running benches/id_creation_benchmarks.rs (target/release/deps/bench_resource_id_creation-69c86f48219418ac)
Gnuplot not found, using plotters backend
Benchmarking resource_id_creation/compute_from_path:../test-assets/lena.jpg: Collecting 100 samples in estimated 20.147 s (232kresource_id_creation/compute_from_path:../test-assets/lena.jpg
                        time:   [86.469 µs 86.533 µs 86.612 µs]
                        change: [-0.9975% -0.6224% -0.3123%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 11 outliers among 100 measurements (11.00%)
  6 (6.00%) high mild
  5 (5.00%) high severe
Benchmarking resource_id_creation/compute_from_path:/Users/tareknasser/Documents/books/Database_System_Concepts.p...: Warming uBenchmarking resource_id_creation/compute_from_path:/Users/tareknasser/Documents/books/Database_System_Concepts.p...: Collectinresource_id_creation/compute_from_path:/Users/tareknasser/Documents/books/Database_System_Concepts.p...
                        time:   [9.5751 ms 9.5894 ms 9.6052 ms]
                        change: [-1.4499% -0.9298% -0.4874%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 7 outliers among 100 measurements (7.00%)
  4 (4.00%) high mild
  3 (3.00%) high severe
```
Performance seems to be unchanged.

kirillt:
We should test build of `data-resource` crate with both `non-cryptographic-hash` and `cryptographic-hash` features on CI. Right now, `cryptographic-hash` is the default. On my machine, `non-cryptographic-hash` fails to build.

tareknaser:
I've added two more tasks to the CI workflow. These tasks check if `data-resource` can be built with both the default feature (`cryptographic-hash`) and `non-cryptographic-hash`. Right now, these tasks are only part of the Linux job.

Once we reach the second milestone as discussed in https://github.com/ARK-Builders/ark-rust/issues/29#issuecomment-2068085981, we can simplify things by using these commands to check compilation features:
```sh
cargo clippy --workspace
```
```sh
cargo clippy --workspace --features non-cryptographic-hash --no-default-features
```
and 
```sh
cargo test --workspace
```
```sh
cargo test --workspace --features non-cryptographic-hash --no-default-features
```
These commands will handle linting and running tests for all relevant crates with different features.

data-resource/src/blake3.rs
kirillt: Thanks! Let's stick to `String` then

kirillt:
So, after we finish #42 we should have approximately this layout:
- `data-resource` will be really light crate, containing only the `ResourceId` trait
- `dev-hash`, or `data-resource-test` defines reference implementations of `ResourceId`: CRC-32 and Blake3, later we'll extend this crate with other functions for benchmarking
- `fs-index` depends on `data-resource` and operates with abstract `ResourceId`
- `fs-storage`, `fs-metadata`, `fs-properties`,etc. ⏤ same
- `ark-cli` depends on `dev-hash` because it's a Swiss Army Knife
- benchmarks depend on `dev-hash`, is it possible in release profile? or only if the benchmarks are in separate crate?

tareknaser:
> - `fs-index` depends on `data-resource` and operates with abstract `ResourceId`
 
But wouldn't that be an issue if `data-resource` only contains `ResourceIdTrait`?  
`fs-index` and other workspace crates still need a `ResourceId` struct to operate.
As an example, in `fs-index`, the hash is a field stored in `ResourceIndex`.
 
> - `ark-cli` depends on `dev-hash` because it's a Swiss Army Knife

What sets `ark-cli` apart from other crates that require `ResourceId`? How does the usage of `ResourceId` differ between `fs-index` and `ark-cli`?

> - benchmarks depend on `dev-hash`, is it possible in release profile? or only if the benchmarks are in separate crate?

We could write benchmarks directly into the crate housing the implementations of `ResourceIdTrait`. This means that the hash function benchmarks would be located within the `dev-hash` crate itself for this proposal.

> Compilation features `cryprographic-hash` and `non-cryptographic-hash` could be moved into `fs-index` because only its code is conditioned

I think it would be useful for the crate containing multiple implementations of `ResourceId` to include compilation flags representing the hash function used for each `ResourceId` implementation. For instance, the `dev-hash` or `data-resource-test` crate could have feature flags to export only the chosen `ResourceId` type.

Here's the reasoning: 
Let's consider that `dev-hash` crate contains seven different implementations of `ResourceIdTrait` for seven different hash functions, some cryptographic and some not. 
As a user, if I only want to use the `blake3` implementation and have no interest in the others, having feature flags would allow me to select only the necessary crate dependencies, to avoid unnecessary bloat.

> actually we can replace 2 flags with 1 flag (e.g. `fast`, or opposite `secure`, or `cryptographic-hash`).

I agree. 
I don't see any rationale for having two feature flags. We could get the same results with just one.

kirillt:
> > fs-index depends on data-resource and operates with abstract ResourceId

> But wouldn't that be an issue if data-resource only contains ResourceIdTrait?
> fs-index and other workspace crates still need a ResourceId struct to operate.
> As an example, in fs-index, the hash is a field stored in ResourceIndex.

I mean something like this:
```
pub trait ResourceId {

}

pub struct IndexEntry<Id: ResourceId> {
    pub modified: SystemTime,
    pub id: Id,
}
```

> > ark-cli depends on dev-hash because it's a Swiss Army Knife

> What sets ark-cli apart from other crates that require ResourceId? How does the usage of ResourceId differ between fs-index and ark-cli?

The difference is that `ark-cli` is an app, not really a crate. Users of ARK framework could use `ark-cli` as a tool, or as a reference to implement something. But they will not depend on `ark-cli` as a crate. We could use only one hash function in `ark-cli`, but it seems that we'll need to cover at least several the most common scenarios of usage. This also means that the hash function would be selectable in `ark-cli` as a CLI parameter.

> > benchmarks depend on dev-hash, is it possible in release profile? or only if the benchmarks are in separate crate?

> We could write benchmarks directly into the crate housing the implementations of ResourceIdTrait. This means that the hash function benchmarks would be located within the dev-hash crate itself for this proposal.

Agree, I've also came to this idea that we should move the benchmarks into `dev-hash`. We can also make `dev-hash` depend on `fs-index` and any other crates, so all benchmarks would be there. In fact, we could just use name `benchmarks` for this 😁 

> > Compilation features cryprographic-hash and non-cryptographic-hash could be moved into fs-index because only its code is conditioned

> I think it would be useful for the crate containing multiple implementations of ResourceId to include compilation flags representing the hash function used for each ResourceId implementation. For instance, the dev-hash or data-resource-test crate could have feature flags to export only the chosen ResourceId type.

The `dev-hash` or `benchmarks` should not export anything. It's more like a reference/mock app. Also, we would need all hash functions for benchmarking. So, compilation flags are not necessary here. By the way, calling this folder `benchmarks` could make it clear that it's not suitable as a dependency for the users. It'd still be possible to use it for quick prototyping 😉

tareknaser:
Moved to #42

PR #33, opened 2024-04-21
Add a Note for Benchmarks' Profiles in `README`

tareknaser:
Resolves https://github.com/ARK-Builders/ark-rust/issues/40
Related issue: ARK-Builders/arklib#30

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,3 +86,13 @@ To generate a flame graph for `index_build_benchmark`, use the following command
 ```bash
 cargo flamegraph --bench index_build_benchmark -o index_build_benchmark.svg -- --bench
 ```
+
+> [!NOTE]
+> Running `cargo flamegraph` on macOS requires `DTrace`. MacOS System Integrity Protection (`SIP`) is enabled by default, which restricts most uses of `dtrace`.
+>
+> To work around this issue, you have two options:
+>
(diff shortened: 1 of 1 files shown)


kirillt:
@tareknaser have you encountered this error?
```
...
    Finished bench [optimized + debuginfo] target(s) in 32.70s

dtrace: system integrity protection is on, some features will not be available
dtrace: failed to initialize dtrace: DTrace requires additional privileges

failed to sample program
```

1. I guess, the privileges problem can be worked around by using `sudo`. Although that's not great solution.
2. What is system integrity in `dtrace`?

tareknaser:
I've noticed the same problem on my Mac. 

Found this information online:
"On all current MacOS versions, System Integrity Protection (`SIP`) is enabled by default and prevents most uses of `dtrace` and other tools and scripts based on it (i.e. `dtruss`)"

kirillt:
@tareknaser so if I understand correctly, benchmarking on macOS is only possible after restarting in recovery mode...

image

kirillt:
@tareknaser could you add a note about this into README, too?

kirillt:
@tareknaser it seems there is another recipe which allows to restart in recovery mode only once: https://forums.developer.apple.com/forums/thread/17452

Could you validate this instruction?

kirillt (approved):

Issue #21, opened 2024-03-19
Filesystem monitoring using `notify-rs`


tareknaser:
## Description 
Use [`notify-rs`](https://github.com/notify-rs/notify) to enable filesystem monitoring for a specified root folder. This functionality would automatically update the index if any changes were detected within the watched directory.
### Motivation
- Enhance UX with a straightforward solution for keeping the index up-to-date.
### Proposal
We've decided to add a method `ResourceIndex::watch_index()` to monitor a specific directory for changes and update the index whenever a change is detected.  
For interacting with this method, we'll add a command to `ark-cli` which is `ark-cli watch <path>`

To further build on this, we plan to enhance this functionality to enable watching multiple directories simultaneously. This will be possible through the following commands:

- `ark-cli watch --all` to watch all specified directories.
- `ark-cli watch pin <path>` to add a directory to the watch list.
- `ark-cli watch unpin <path>` to stop watching a directory.
- `ark-cli watch list` to list the directories currently being watched.

We can also have a global configuration file, such as `/Users/<username>/.ark/directories_to_watch`, to store the list of directories to monitor.

PR #36, opened 2024-04-22
Add `watch_index` method and `ark-cli watch` command

tareknaser:
## Description
This pull request adds a new method to `fs-index` crate, `watch_index`, to monitor file system changes and automatically update the index. 
Additionally, it adds a new command to `ark-cli` to make this functionality accessible to users. 
This change is the first step of addressing issue #21.
## Testing
An example of the new method's usage is in the `fs-index` crate at `fs-index/examples/index_watch.rs`. 
To run the example, run the following command:
```sh
cargo run --example index_watch
```
This command monitors the index at the `test-assets/` directory and automatically updates it upon any file system changes.

### DIFF
--- a/fs-index/Cargo.toml
+++ b/fs-index/Cargo.toml
@@ -14,6 +14,12 @@ walkdir = "2.3.2"
 anyhow = "1.0.58"
 serde_json = "1.0"
 serde = { version = "1.0", features = ["derive"] }
+# For the watch API
+notify = { version = "6.1", optional = true }
+notify-debouncer-full = { version = "0.3", optional = true }
+futures = { version = "0.3", optional = true }
+async-stream = { version = "0.3", optional = true }
+tokio = { version = "1.40", features = ["full"], optional = true }
 
 
 fs-storage = { path = "../fs-storage" }
@@ -21,8 +27,13 @@ fs-storage = { path = "../fs-storage" }
 data-error = { path = "../data-error" }
 data-resource = { path = "../data-resource" }
 
+
+[features]
+watch = ["notify", "notify-debouncer-full", "futures", "async-stream", "tokio"]
+
 [dev-dependencies]
 uuid = { version = "1.6.1", features = ["v4"] }
+env_logger = "0.11"
 # benchmarking
 criterion = { version = "0.5", features = ["html_reports"] }
 tempfile = "3.10"
--- a/fs-index/examples/index_watch.rs
+++ b/fs-index/examples/index_watch.rs
@@ -1,0 +1,35 @@
+use std::path::Path;
+
+use anyhow::Result;
+use futures::{pin_mut, StreamExt};
+
+use dev_hash::Blake3;
+use fs_index::{watch_index, WatchEvent};
+
+/// A simple example of using `watch_index` to monitor a directory for file
+/// changes. This asynchronously listens for updates and prints the paths of
+/// changed files.
+#[tokio::main]
+async fn main() -> Result<()> {
+    env_logger::init();
+
+    // Change this to the path of the directory you want to watch
+    let root = Path::new("test-assets");
+
+    let stream = watch_index::<_, Blake3>(root);
+
+    pin_mut!(stream); // needed for iteration
+
+    while let Some(value) = stream.next().await {
+        match value {
+            WatchEvent::UpdatedOne(update) => {
+                println!("Updated file: {:?}", update);
+            }
+            WatchEvent::UpdatedAll(update) => {
+                println!("Updated all: {:?}", update);
+            }
+        }
+    }
+
+    Ok(())
+}
--- a/fs-index/src/index.rs
+++ b/fs-index/src/index.rs
@@ -65,6 +65,17 @@ pub struct Timestamped<Item> {
     pub(crate) last_modified: SystemTime,
 }
 
+impl<Item> Timestamped<Item> {
+    pub fn item(&self) -> &Item {
+        &self.item
+    }
+
+    /// Return the last modified time
+    pub fn last_modified(&self) -> SystemTime {
+        self.last_modified
+    }
+}
+
 type IndexedPaths = HashSet<Timestamped<PathBuf>>;
 
 /// Represents the index of resources in a directory.
@@ -486,14 +497,21 @@ impl<Id: ResourceId> ResourceIndex<Id> {
     /// - The index is up-to-date with the file system except for the updated
     ///   resource
     /// - In case of a addition, the resource was not already in the index
-    /// - In case of a modification or removal, the resource was already in the
-    ///   index
+    /// - In case of a removal, the resource was already in the index
+    /// - In case of a move or rename, `update_one()` should be called twice:
+    ///   once with the old path to remove the previous entry, and once with the
+    ///   new path to add the updated entry
     pub fn update_one<P: AsRef<Path>>(
         &mut self,
         relative_path: P,
-    ) -> Result<()> {
+    ) -> Result<IndexUpdate<Id>> {
         let path = relative_path.as_ref();
         let entry_path = self.root.join(path);
+
+        let mut result = IndexUpdate {
+            added: HashMap::new(),
+            removed: HashSet::new(),
+        };
 
         // Check if the entry exists in the file system
         if !entry_path.exists() {
@@ -505,16 +523,18 @@ impl<Id: ResourceId> ResourceIndex<Id> {
                 "Caller must ensure that the resource exists in the index: {:?}",
                 path
             );
-            let id = self.path_to_id.remove(path).unwrap();
+            let id = self.path_to_id.remove(path).ok_or_else(
+                || anyhow::anyhow!("Resource with path {} is neither in the index nor in the file system. Make sure the index is up-to-date.", path.display())
+            )?;
             self.id_to_paths
                 .get_mut(&id.item)
-                .unwrap()
+                .expect("Resource ID not found in the ID to paths map")
                 .remove(path);
             // If the ID has no paths, remove it from the ID to paths map
             if self.id_to_paths[&id.item].is_empty() {
                 self.id_to_paths.remove(&id.item);
+                result.removed.insert(id.item);
             }
-
             log::trace!("Resource removed: {:?}", path);
         } else {
             // If the entry exists in the file system, it's an addition or
@@ -534,7 +554,7 @@ impl<Id: ResourceId> ResourceIndex<Id> {
             if let Some(prev_id) = self.path_to_id.get(path) {
                 self.id_to_paths
                     .get_mut(&prev_id.item)
-                    .unwrap()
+                    .expect("Resource ID not found in the ID to paths map")
                     .remove(path);
             }
 
@@ -547,9 +567,16 @@ impl<Id: ResourceId> ResourceIndex<Id> {
                 .or_default()
                 .insert(path.to_path_buf());
 
+            let timpestamped_path = Timestamped {
+                item: path.to_path_buf(),
+                last_modified,
+            };
+            result
+                .added
+                .insert(id, HashSet::from([timpestamped_path]));
             log::trace!("Resource added/updated: {:?}", path);
         }
 
-        Ok(())
+        Ok(result)
     }
 }
--- a/fs-index/src/tests.rs
+++ b/fs-index/src/tests.rs
@@ -810,7 +810,12 @@ fn test_track_removal_with_collision() {
 
         fs::remove_file(&file_path).expect("Failed to remove file");
 
-        index.update_one("file1.txt").expect("Failed to update index");
+        let result = index.update_one("file1.txt").expect("Failed to update index");
+
+        // Assert that `update_one` result is empty
+        // Rational: There is still a file with the same content as file1.txt so the
+        //           resource was not removed.
+        assert_eq!(result.removed().len(), 0, "{:?}", result);
 
         assert_eq!(index.len(), 1, "{:?}", index);
         let resource_by_path = index
@@ -906,5 +911,48 @@ fn test_track_modification_with_collision() {
         // The length of `collisions` should be 1 because file1.txt and file2.txt
         // have the same content.
         assert_eq!(index.collisions().len(), 1, "{:?}", index);
+    });
+}
+
+/// Test for calling `update_one()` on a file that was moved from the root
+/// directory to a subdirectory.
+///
+/// ## Test scenario:
+/// - Create a file within the temporary directory.
+/// - Build a resource index in the temporary directory.
+/// - Move the file to a subdirectory.
+/// - Call `update_one()` 2 times with the relative path of the moved file.
+/// - Assert that the index contains the expected number of entries with the
+///  correct IDs and paths after the move.
+#[test]
+fn test_track_move_to_subdirectory() {
+    for_each_type!(Crc32, Blake3 => {
+        let temp_dir = TempDir::with_prefix("ark_test_track_move_to_subdirectory")
+            .expect("Failed to create temp dir");
+        let root_path = temp_dir.path();
+
+        let file_path = root_path.join("file.txt");
+        fs::write(&file_path, "file content").expect("Failed to write to file");
+        let file_id = Id::from_path(&file_path).expect("Failed to get checksum");
+
+        let mut index: ResourceIndex<Id> =
+            ResourceIndex::build(root_path).expect("Failed to build index");
+
+        let subdirectory_path = root_path.join("subdirectory");
+        fs::create_dir(&subdirectory_path).expect("Failed to create subdirectory");
+
+        let moved_file_path = subdirectory_path.join("file.txt");
+        fs::rename(&file_path, &moved_file_path).expect("Failed to move file");
+
+        // We need to call `update_one()` 2 times because the file was moved to a
+        // subdirectory.
+        index.update_one("file.txt").expect("Failed to update index");
+        index.update_one("subdirectory/file.txt").expect("Failed to update index");
+
+        assert_eq!(index.len(), 1, "{:?}", index);
+        let resource_by_path = index
+            .get_resource_by_path("subdirectory/file.txt")
+            .expect("Failed to get resource");
+        assert_eq!(*resource_by_path.id(), file_id);
     });
 }
--- a/fs-index/src/watch.rs
+++ b/fs-index/src/watch.rs
@@ -1,0 +1,169 @@
+use std::{fs, path::Path, thread, time::Duration};
+
+use async_stream::stream;
+use futures::Stream;
+use notify::{RecursiveMode, Watcher};
+use notify_debouncer_full::new_debouncer;
+use tokio::sync::mpsc;
+
+use data_resource::ResourceId;
+use fs_storage::ARK_FOLDER;
+
+use crate::{IndexUpdate, ResourceIndex};
+
+/// Represents the different kinds of events that can occur when watching the
+/// resource index.
+#[derive(Debug)]
+pub enum WatchEvent<Id: ResourceId> {
+    /// Represents an update to a single resource.
+    UpdatedOne(IndexUpdate<Id>),
+    /// Represents an update to all resources.
+    UpdatedAll(IndexUpdate<Id>),
+}
+
+/// Watches for file system changes and emits events related to the
+/// [ResourceIndex].
+///
+/// This function sets up a file watcher that monitors a specified root path for
+/// changes to files. It sends events such as file creations, modifications,
+/// renames, and deletions through an asynchronous stream. The function uses a
+/// debouncer to ensure that multiple rapid events are collapsed into a single
+/// event.
+pub fn watch_index<P: AsRef<Path>, Id: ResourceId + 'static>(
+    root_path: P,
+) -> impl Stream<Item = WatchEvent<Id>> {
+    log::debug!(
+        "Attempting to watch index at root path: {:?}",
+        root_path.as_ref()
+    );
+
+    let root_path = fs::canonicalize(root_path.as_ref()).unwrap();
+    let mut index: ResourceIndex<Id> =
+        ResourceIndex::build(&root_path).unwrap();
+    index.store().unwrap();
+
+    let (tx, mut rx) = mpsc::channel(100);
+    let ark_folder = root_path.join(ARK_FOLDER);
+
+    // We need to spawn a new thread to run the blocking file system watcher
+    thread::spawn(move || {
+        // Setup the synchronous channel (notify debouncer expects this)
+        let (sync_tx, sync_rx) = std::sync::mpsc::channel();
+
+        let mut debouncer =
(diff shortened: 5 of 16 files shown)


fs-index/examples/index_watch.rs
kirillt, line 1: Do we need both example and CLI command?

tareknaser: I've added a simple usage example to the crate to showcase its basic functionality. The CLI command would eventually grow in complexity with new features.

Do you prefer that I remove the example from `fs-index`?

kirillt: No, actually I think it's good.

fs-index/src/watch.rs
```diff
@@ -0,0 +1,69 @@
...
+                info!("Detected event: {:?}", event);
+                index.update_all()?;
+                index.store()?;
+                info!("Index updated and stored");
+            }
+            Err(e) => log::error!("Error in watcher: {:?}", e),
+        }
+    }
+
+    Ok(())
```
kirillt, line 50: Can this statement be reached at all?

Maybe should be `!unreachable!()`?

fs-index/src/watch.rs
```diff
@@ -0,0 +1,69 @@
...
+    let config = Config::default();
+    watcher.configure(config)?;
+    watcher.watch(root_path.as_ref(), RecursiveMode::Recursive)?;
+    info!("Started watcher with config: \n\t{:?}", config);
+
+    while let Some(res) = rx.next().await {
+        match res {
+            Ok(event) => {
+                // If the event is a change in .ark folder, ignore it
+                let ark_folder = root_path.join(ARK_FOLDER);
```
kirillt, line 32: Is it worth moving out of the loop?

fs-index/src/watch.rs
```diff
@@ -0,0 +1,69 @@
...
+                if event
+                    .paths
+                    .iter()
+                    .any(|p| p.starts_with(&ark_folder))
+                {
+                    continue;
+                }
```
kirillt, line 39: I would expect something like this:
```
let paths = event.paths.filter(|p| !p.start_with(&ark_folder));

index.update(paths)
```

fs-index/src/watch.rs
```diff
@@ -0,0 +1,69 @@
...
+    let mut index = ResourceIndex::provide(&root_path)?;
+
+    let (mut watcher, mut rx) = async_watcher()?;
+    info!("Watching directory: {:?}", root_path);
+    let config = Config::default();
+    watcher.configure(config)?;
+    watcher.watch(root_path.as_ref(), RecursiveMode::Recursive)?;
+    info!("Started watcher with config: \n\t{:?}", config);
+
+    while let Some(res) = rx.next().await {
```
kirillt, line 28: From docs for `notify::event::Event`:
```
You might want to check Event::need_rescan to make sure no event was missed before you received this one.
```

This sounds like we should match `Event::need_rescan` and perform `update_all` in this case.

fs-index/src/watch.rs
kirillt: So, we need new `fn update(path: Vec<Path>)` for this. It would be also called by `update_all`.

fs-index/src/watch.rs
```diff
@@ -0,0 +1,69 @@
...
+                if event
+                    .paths
+                    .iter()
+                    .any(|p| p.starts_with(&ark_folder))
+                {
+                    continue;
+                }
+
+                info!("Detected event: {:?}", event);
+                index.update_all()?;
```
kirillt, line 42: This looks too inefficient. If we call `update_all` for each update, why not just call it once in a while? I think we should take changed paths, filter out those starting with `.ark`, then update index only using the modified paths.

kirillt:
It's a good PR, and it seems to be pretty straightforward to complete it, but I'm afraid that merging it before the other index refactorings could make porting https://github.com/ARK-Builders/arklib/pull/72 too difficult. Because we'd need to add one more function to the index, and at the same time we need to check diffs while porting https://github.com/ARK-Builders/arklib/pull/72.

fs-index/src/watch.rs
tareknaser: I can't think of a case where we reach this statement except maybe if the `rx` channel closes for some reason.
Updated

fs-index/src/watch.rs
tareknaser: Yes :)
Thanks for the catch

fs-index/src/watch.rs
tareknaser: Exactly. I'll implement a helper function to update the indexes of multiple directories with one call. This will be part of the next phase, which will be handled in a separate PR. 

Specifically, we'll require this function for the 
```sh
ark-cli watch --all
```
command

fs-index/src/watch.rs
tareknaser: Thanks. I didn't notice

fs-index/src/watch.rs
tareknaser: You're right. 
I went through the docs for the event types
```rust
pub enum EventKind {
    Any,
    Access(AccessKind),
    Create(CreateKind),
    Modify(ModifyKind),
    Remove(RemoveKind),
    Other,
}
```

I suggest pausing the changes in this PR until we update the ResourceIndex API as discussed in [https://github.com/ARK-Builders/arklib/issues/91#issuecomment-1998887042](https://github.com/ARK-Builders/arklib/issues/91#issuecomment-1998887042)

Then, we can follow this workflow:
- Check for `need_rescan()`. If true, call `update_all()`.
- Filter out events in `.ark/`.
- Use a match statement for the type of event:
    - `Create` -> call `track_addition` with the relative path.
    - `Modify` -> call `track_update`.
    - `Remove` -> call `track_deletion`.

These actions seem to align perfectly with the methods we agreed on for the API.

kirillt: Cool that the "Track API" appeared to be useful already.

By the way, we might utilize `Access(AccessKind)`, too. Like providing "recently viewed" feature, which would take into account all users devices. We should design this feature later.

tareknaser:
Updated the watch API to call `ResourceIndex::update_one()` for files created, removed, or modified based on streams from `notify` events.

tareknaser:
There appear to be some unexpected events coming from the `notify` stream. For example, I've identified a potential flaw with the following steps:
1. Run the watch API on a folder.
2. Copy a file multiple times (e.g., `file copy.txt`, `file copy 2.txt`).
3. Up until this point, the index updates correctly.
4. Delete both files simultaneously (select and delete them together).
This results in a panic in `ResourceIndex::update_one()`.

This situation requires further investigation. Additionally, we need to test this scenario alongside other `ResourceIndex` tests to be implemented for #88.

kirillt:
@tareknaser does only simultaneous deletion cause problems? Does simultaneous addition work fine?

fs-index/src/watch.rs
kirillt: It seems that handling `Event::need_rescan` is still missing.

fs-index/src/watch.rs
```diff
@@ -0,0 +1,157 @@
...
+                            kind:
+                                notify::EventKind::Remove(
+                                    notify::event::RemoveKind::File,
+                                ),
+                            ..
+                        } => {}
+                        _ => continue,
+                    },
+                    _ => {}
+                }
```
kirillt, line 108: Can't we match `event` directly without nested `match` expressions?

tareknaser:
> does only simultaneous deletion cause problems? Does simultaneous addition work fine?

Yes and yes 
Even simultaneous deletion work fine in some cases but i was able to reproduce the error more than once

fs-index/src/watch.rs
tareknaser: Yes, I’m not sure why I wrote it that way in the first place 😃 

Updated

kirillt:
README should be updated to explicitly state in which folder this command should be run:
```
cargo run --example resource_index
```

Also, would be great to adjust the example to allow passing arbitrary path.

Or, if `ark-cli` can handle same scenario, it should be mentioned in the README.

fs-index/src/watch.rs
```diff
@@ -0,0 +1,132 @@
...
+/// directory tree. Events related to the internal `.ark` folder are ignored to
+/// prevent unnecessary updates.
+///
+/// # Arguments
+///
+/// * `root_path` - The root directory to be watched. This path is canonicalized
+///   to handle symbolic links and relative paths correctly.
+pub async fn watch_index<P: AsRef<Path>, Id: ResourceId>(
+    root_path: P,
+) -> Result<()> {
```
kirillt, line 35: This public function seems is the only API entrypoint, but in fact developer cannot extract any information from running the watcher. How would a developer integrate with Watch API?

I think, we need some kind of stream or events emitter so the developer would be able to handle the indexing events. The content of the stream should be same as output of `update_all`.

fs-index/src/watch.rs
```diff
@@ -0,0 +1,139 @@
...
+                        notify::event::RemoveKind::File,
+                    ) => {}
+                    _ => continue,
+                }
+
+                // If the event requires a rescan, update the entire index
+                // else, update the index for the specific file
+                if event.need_rescan() {
+                    info!("Detected rescan event: {:?}", event);
+                    index.update_all()?;
```
kirillt, line 89: The result should be used to provide user with actual updates.

fs-index/src/watch.rs
```diff
@@ -0,0 +1,139 @@
...
+                            .resources()
+                            .iter()
+                            .map(|x| x.path().to_str().unwrap().to_string())
+                            .collect::<Vec<String>>()
+                            .join("\n\t")
+                    );
+
+                    let relative_path = file.strip_prefix(&root_path)?;
+                    log::info!("Relative path: {:?}", relative_path);
+                    index.update_one(relative_path)?;
```
kirillt, line 110: The result should be used to provide user with actual updates.

tareknaser:
I spent some time today looking into different ways to use `notify` by going through the docs and [examples](https://github.com/notify-rs/notify/tree/main/examples). Right now, we're using `async_monitor`, but it might not be the best choice for us.

Order is very important in our case because we need events to happen in the right order (for example, we don’t want to see "`file1.txt` removed" before "`file1.txt` created", since this would mess up our `update_one()` logic). Using an asynchronous watcher could cause issues with keeping the events in order.

> Btw, I think this might be why we saw [this error](https://github.com/ARK-Builders/ark-core/pull/36#issuecomment-2337822877). As I reported, the error wasn’t consistent, which could be because asynchronous events don't always happen in the right order.

While looking through the examples, I also noticed that we might want to use the Debouncer. [The file system can sometimes send multiple events for what is really just one change](https://stackoverflow.com/a/1765094), which could cause problems. For example, it might trigger `update_one()` several times when a file is created.

I'm now testing this in a smaller example and looking at how to set up the event stream properly.

fs-index/src/watch.rs
tareknaser: Updated `watch_index()` to return a stream of `WatchEvent<Id>`, where `WatchEvent` is:
```rust
/// Represents the different kinds of events that can occur when watching the
/// resource index.
#[derive(Debug)]
pub enum WatchEvent<Id: ResourceId> {
    /// Represents an update to a single resource.
    UpdatedOne(IndexUpdate<Id>),
    /// Represents an update to all resources.
    UpdatedAll(IndexUpdate<Id>),
}
```
This contains information about changes, whether it involves a single updated resource (`UpdatedOne`) or all resources (`UpdatedAll`).

fs-index/src/watch.rs
tareknaser: I chose the first approach to get things running faster and keep it simpler for now. I also plan to add tests soon.

Next, I want to add integration tests for this functionality. We could either add integration tests for `fs-index` directly or implement CI shell scripts to test `ark-cli watch <PATH>` for an end-to-end approach—possibly both?

Do you think CI shell scripts for `ark-cli watch` would be sufficient, or should we also include programmatic tests? For example, running the watcher in a separate thread and doing many create/delete operations to verify the results. Now that I think about it, writing these tests programmatically could be complex. What’s your take?

fs-index/Cargo.toml
```diff
@@ -10,10 +10,14 @@ bench = false
...
 [dependencies]
 log = { version = "0.4.17", features = ["release_max_level_off"] }
+env_logger = "0.11"
 walkdir = "2.3.2"
 anyhow = "1.0.58"
 serde_json = "1.0"
 serde = { version = "1.0", features = ["derive"] }
+# For the watch API
+notify = "6.1"
+futures = "0.3"
```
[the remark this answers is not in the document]
tareknaser: Pushed these behind a feature flag and included a note in the README for it as well.

fs-index/src/watch.rs
kirillt: > Do you think CI shell scripts for ark-cli watch would be sufficient, or should we also include programmatic tests? For example, running the watcher in a separate thread and doing many create/delete operations to verify the results. Now that I think about it, writing these tests programmatically could be complex. What’s your take?

Agree, I think we can achieve proper result by simple shell script. I imagine it like this:
1. Run `ark-cli watch` in background, direct its output to dedicated log file.
2. Randomly modify folder content and write the performed actions into another log file.
3. Then compare the two log files.

fs-index/src/index.rs
```diff
@@ -65,6 +65,17 @@ pub struct Timestamped<Item> {
...
+impl<Item> Timestamped<Item> {
+    pub fn item(&self) -> &Item {
+        &self.item
+    }
+
+    /// Return the last modified time
+    pub fn last_modified(&self) -> SystemTime {
+        self.last_modified
+    }
+}
```
kirillt, line 77: Just curious, why do we need this getters?

fs-index/src/index.rs
```diff
@@ -515,6 +531,7 @@ impl<Id: ResourceId> ResourceIndex<Id> {
                 self.id_to_paths.remove(&id.item);
             }
 
+            result.removed.insert(id.item);
```
kirillt, line 534: So, if I'm not mistaken, this line should be moved under the condition above:
```
if self.id_to_paths[&id.item].is_empty() {
    self.id_to_paths.remove(&id.item);
    // emit the removed event
}
```

I.e. only removal of the last path with same id should be considered real removal.

And we could log duplicates removal in `else` branch of same condition.

fs-index/src/index.rs
```diff
@@ -547,9 +564,16 @@ impl<Id: ResourceId> ResourceIndex<Id> {
                 .or_default()
                 .insert(path.to_path_buf());
 
+            let timpestamped_path = Timestamped {
+                item: path.to_path_buf(),
+                last_modified,
+            };
+            result
+                .added
+                .insert(id, HashSet::from([timpestamped_path]));
```
kirillt, line 573: We could wrap this into similar condition, checking that the amount of paths mapped to the id is zero. And log that duplicate was introduced in the corresponding `else` branch.

This should allow us relax this requirement to `update_one`:
```
    /// - In case of a addition, the resource was not already in the index
```

kirillt:
Nitpick, but we can define aliases `IndexUpdate::addition(id, path)` and `IndexUpdate::removal(id)` for these snippets:
```
result.removed.insert(id.item);
```

```
result
    .added
    .insert(id, HashSet::from([timpestamped_path]));
```

Then we could immediately return from `update_one` once we determined the update:
```
return Ok(IndexUpdate::removal(id.item));
```
```
return Ok(IndexUpdate::addition(id, timpestamped_path));
```
This should be more readable.

----

By the way, it seems that we could simplify `added` field of the `IndexUpdate` structure. Since we don't distinguish duplicates, we can take any path as a representative of the group, so the app could do something with it. In practice, when unique resource is detected, we take its path as the representative. When a duplicate appears, we skip it. If during unique addition, several paths were introduced at once, we take an arbitrary one (options: 1) random; 2) just first in the vector; 3) the shortest path).

From API point of view, we don't need a collection of paths attached to the addition event, only one path (representative).

----

We might need separate events to track duplicates. Something like `DuplicateAdded(id, path)` and `DuplicateRemoved(id, path)`. Although, I'm not sure that duplicate removal can be useful, maybe duplicate addition is enough. It could be used to allow the user to select representative manually. Just idea for future.

fs-index/src/index.rs
tareknaser: They’re just there to access `Timestamped` fields without accidentally messing with them. Like, if I do calculations on `last_modified`, I don’t want to change the original value by mistake.

fs-index/src/index.rs
tareknaser: This one’s actually different since `IndexUpdate::added` is of type `HashMap<Id, HashSet<Timestamped<PathBuf>>>`, so we need to specify the path of the added resources.

tareknaser:
> Then we could immediately return from `update_one` once we determined the update:

`update_one()` currently has an if statement that checks early on if the update is a removal or an addition, which makes it able to return from each branch as soon as possible.


> By the way, it seems that we could simplify `added` field of the `IndexUpdate` structure.

I was actually looking at the code today and thought the same thing. The added field definitely has an interesting type, though I can’t quite remember why we landed on it—probably something inherited from older `arklib` code.

I think we could track this in a separate issue and handle it in its own PR. What do you think?

fs-index/src/watch.rs
tareknaser: Added a shell script `integration/ark-cli-watch.sh` to check the sanity of `ark-cli watch`. I think it’s pretty cool because it also checks other parts of the code for sanity along the way, like `update_one` in this case. Using `ark-cli` is definitely a great way to write end-to-end shell scripts to verify what we have.
  
I’ve also set it up to run in the CI with each push/PR to the main branch. You can check out the expected workflow in [my fork here](https://github.com/tareknaser/ark-rust/actions/runs/11576918354). I’ve been using it for debugging

fs-index/src/index.rs
```diff
@@ -529,9 +529,8 @@ impl<Id: ResourceId> ResourceIndex<Id> {
             // If the ID has no paths, remove it from the ID to paths map
             if self.id_to_paths[&id.item].is_empty() {
                 self.id_to_paths.remove(&id.item);
+                result.removed.insert(id.item);
             }
-
-            result.removed.insert(id.item);
```
kirillt: We need some test cases for this.

As well as test cases demonstrating what happens if the caller violates assumptions:
>     /// **Note**: The caller must ensure that:
>    /// - The index is up-to-date with the file system except for the updated
>    ///   resource
>    /// - In case of a addition, the resource was not already in the index
>    /// - In case of a modification or removal, the resource was already in the
>    ///   index

I'm thinking now that we should panic when the assumptions are violated.

integration/ark-cli-watch.sh
kirillt, line 1: Good idea. Could you record short screencast using it, after we merge it?

kirillt:
> > Then we could immediately return from `update_one` once we determined the update:

> `update_one()` currently has an if statement that checks early on if the update is a removal or an addition, which makes it able to return from each branch as soon as possible.

I'm talking only about code clarity/maintainability.

kirillt:
> I was actually looking at the code today and thought the same thing. The added field definitely has an interesting type, though I can’t quite remember why we landed on it—probably something inherited from older `arklib` code.

We can have multiple paths for same id, and we can have many ids when use `update_all`.

> I think we could track this in a separate issue and handle it in its own PR. What do you think?

Agree, created issue:
- #93

integration/ark-cli-watch.sh
tareknaser: sure. created an asana follow up task to keep track of this

fs-index/src/index.rs
```diff
@@ -491,10 +502,15 @@ impl<Id: ResourceId> ResourceIndex<Id> {
     pub fn update_one<P: AsRef<Path>>(
```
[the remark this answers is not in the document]
kirillt: Also, I'm not sure that we still require that added resource isn't in the index yet. What happens if it's already tracked? Do we silently consider it modification and update its timestamp? We could add assertion just to prevent mis-use.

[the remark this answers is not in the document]
tareknaser: > It seems, that modification of content by same path doesn't impose any invariants to check before the call.

Exactly. If the developer calls `update_one()` expecting to modify a file that isn’t actually in the index, it will simply treat it as a new addition.


> what happens if this sequence is not followed

At any point in time, if `update_one()` is called with a path that doesn’t exist either in the file system or in the index, it will panic. I noticed this while implementing the file system watcher without the debouncer; it picks up a lot of noisy events, and calling `update_one()` on them causes these panics

>  What happens if it's already tracked?

If the file is already tracked and `update_one()` is called, it’s assumed to be a modification. The file will simply be rescanned and updated in the index without changing its timestamp, since there’s no actual update to the file. 
In other words, calling `update_one()` multiple times on a resource that exists in both the file system and the index has no real impact or downside.

tareknaser:
I made a few updates to the PR:
- Added logging for the `index_watch` example
- Added a note in the `update_one()` doc comment to specify that for rename or move operations, `update_one()` should be called twice
- Added a couple of tests to cover these two cases

fs-index/src/tests.rs
```diff
@@ -913,3 +913,84 @@ fn test_track_modification_with_collision() {
...
+///
+/// ## Test scenario:
+/// - Create a file within the temporary directory.
+/// - Build a resource index in the temporary directory.
+/// - Move the file to a subdirectory.
+/// - Call `update_one()` 2 times with the relative path of the moved file.
+/// - Assert that the index contains the expected number of entries with the
+///  correct IDs and paths after the move.
+#[test]
+fn test_track_move_to_subdirectory() {
```
kirillt, line 928: Is this test-case really different from `fn test_track_rename()`?

Could you create test-case for modification? Btw I'm not sure that 2 calls are needed for modification, because path doesn't change. Let's double-check and correct the comment if needed.

fs-index/src/index.rs
tareknaser: > This is good, but even better would be to provide error message. User could forget to call `update_one` when resource was introduced, but make the call when it was removed.

Updated the code to gracefully return an error if this case happens

fs-index/src/tests.rs
tareknaser: > Is this test-case really different from `fn test_track_rename()`?

It’s not really necessary, but when I was working on this initially, those were the cases where I ran into errors and had to debug for a bit 😃. So, I decided to explicitly include them in the tests. But yes, you’re right—they are essentially the same thing. Let’s just keep the `test_track_move_to_subdirectory` test.

> Could you create test-case for modification?

We already have `test_track_modification` and `test_track_modification_with_collision`. Do you think we need another one?

> Btw I'm not sure that 2 calls are needed for modification, because path doesn't change.

Yes, only one call to `update_one` is needed.

kirillt (approved):
Thanks, great job

PR #41, opened 2024-05-04
Update `FileStorage` write to disk format to valid JSON

### LINKED ISSUE #38: `fs-storage`: Update FileStorage format
We have mixed format now:
* `version 2` came from the legacy plain-text format
* the mapping itself is presented as JSON

```
$ cargo run --example cli write /tmp/test a:1,b:2

$ cat /tmp/test | jq
version 2
{"a":"1","b":"2"}
```

This makes the format not so ergonomic, power users need to perform extra step to remove the header:
```
$ tail -1 /tmp/test | jq 
{
  "a": "1",
  "b": "2"
}
```

It's better to complete transition to JSON.

### Option 1: "Flat syntax"
If we declare `version` in the mapping itself, we'd need to handle it as a special field before reading the entries:
```
{
    "a":"1",
    "b":"2",
    "version":"3"
}
```

Can it cause problems when we expect non-string keys in the storage? E.g. with the following storage:
```
{
    "1":"a",
    "2":"b",
    "version":"3"
}
```

### Option 2: "Nested syntax"
We can use the nested JSON structure, where top-level always consists of 2 fields: `version`, and `entries`. The latter lists all key-values inside:
```
{
    "version":"3",
    "entries": {
        "1":"a",
        "2":"b"
    }
}
```

This format is explicit, no type inconsistencies but the format is more verbose.

tareknaser:
## Description
This pull request addresses the issue where the write to disk format for `FileStorage` was not valid JSON:
- #38

## Changes
Previously, the format looked like this:
```
version 2
{"a":"1","b":"2"}
```
After this pull request, the format will be updated to the following valid JSON format:
```
{
  "version": 3,
  "data": {
    "a": "1",
    "b": "2",
    "c": "3"
  }
}
```
In addition to fixing the format, this pull request also bumps the version to 3 (JSON).

Furthermore, support for reading the legacy version 2 plaintext `FileStorage` format is added for backwards compatibility. This is achieved through the addition of a helper function `read_version_2_fs`, along with a unit test.

### DIFF
--- a/fs-storage/src/file_storage.rs
+++ b/fs-storage/src/file_storage.rs
@@ -1,5 +1,6 @@
+use serde::{Deserialize, Serialize};
 use std::fs::{self, File};
-use std::io::{BufRead, BufReader, BufWriter, Write};
+use std::io::{BufWriter, Write};
 use std::time::SystemTime;
 use std::{
     collections::BTreeMap,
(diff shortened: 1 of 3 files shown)


fs-storage/src/file_storage.rs
```diff
@@ -1,92 +1,101 @@
...
-            data: BTreeMap::new(),
+            data: FileStorageData {
+                version: STORAGE_VERSION,
+                data: BTreeMap::new(),
+            },
         };
 
         // Load the data from the file
-        file_storage.data = match file_storage.read_fs() {
+        file_storage.data.data = match file_storage.read_fs() {
```
kirillt, line 70: What if we rename the inner `data` field into `entries` or `values`?

fs-storage/src/file_storage.rs
```diff
@@ -110,32 +119,50 @@ where
...
-                    let line = line?;
-                    if line.is_empty() {
-                        continue;
-                    }
-                    data.push_str(&line);
-                }
-                let data: BTreeMap<K, V> = serde_json::from_str(&data)?;
-                self.timestamp = new_timestamp;
-                Ok(data)
+        // Attempt to parse the file using the legacy version 2 storage format of FileStorage.
```
kirillt, line 129: Can we scan version number first without actually parsing the data field? And then parse the data according to what the version number is?

kirillt: Something like this:
1. Read the first line.
2. If it's `version 2`, then call `read_version_2_fs`. If it fails, we don't attempt parsing with v3 syntax.
3. If the first line is not `version 2`, then we attempt parsing with v3 syntax.

fs-storage/src/utils.rs
kirillt, line 1: What if we rename this file into `src/v2/parse.rs` with 2 methods:
- `read_header`
- `read_entries`

I'm not sure if we need `v3` module though, I guess it would be empty because reading/writing is done automatically by `serde-json`? If we won't have `v3` module, then we could rename this file into `v2.rs`. But we should read header and entries in separate functions anyway.

fs-storage/src/file_storage.rs
tareknaser: Renamed to `entries`

fs-storage/src/file_storage.rs
tareknaser: > Can we scan version number first without actually parsing the data field?

Sure that makes more sense. Modified the code to fix this issue.

fs-storage/src/utils.rs
kirillt: Good point about FileStorage. Let's keep things simple, no further fixes needed.

PR #61, opened 2024-05-14
JNI bindings for FileStorage

twitu:
Add JNI bindings for FileStorage to create an instance and for all BaseStorage functions.

Once a FileStorage instance is created, it is boxed and leaked as jlong. This jlong can then be upcasted back to a FileStorage instance. All BaseStorage functions take the jlong as an argument upcast it and then perform the necessary operations.

Note: FileStorage needs to be specialized to use with JNI. After discussion, I've specialized it to `FileStorage<String, String>`, however it can be easily changed.

TODO:
* Store the `jlong` in the class instance itself rather than returning it. It will reduce the amount of Java/Kotlin wrapper code that needs to be written
* Handle Results better - Should they return a default value or should they thrown an exception? open to discussion
* Support `read_fs`? It returns the key value pairs as a BTreeMap, however jni by default does not know how to convert BTreeMap to Java's LinkedHashMap. Do we want to support the `read_fs` operation for JNI bindings? This means we'll have to implement a custom conversion logic from BTreeMap to LinkedHashMap


twitu:
The function `Java_FileStorage_read_fs` implements the conversion from `BTreeMap` to `LinkedHashMap`. Going from non-trivial Rust container types to Java container types requires a lot of conversion logic. It will be good to benchmark and decide if it's worth the complexity.

kirillt:
@twitu thanks, looks like a good start!

> Note: FileStorage needs to be specialized to use with JNI. After discussion, I've specialized it to FileStorage<String, String>, however it can be easily changed.

Type `String` for both keys and values sounds good. Probably we'll change it to bytes later, but after benchmarking.

> Store the jlong in the class instance itself rather than returning it. It will reduce the amount of Java/Kotlin wrapper code that needs to be written

I think we shouldn't mix JNI code into pure Rust structures. Desktop developers should not be affected by JNI. Building and importing `fs-storage` alone should not trigger building or importing JNI code. Actually, we need to move all JNI stuff into separate crate `fs-storage-jni`, or even directly to [ark-android](https://github.com/ARK-Builders/ark-android).

> Handle Results better - Should they return a default value or should they thrown an exception? open to discussion

Is there a way to reconstruct Result/Option type? Or would it be too inefficient?

> Support read_fs?

Yes, but converting whole structures really sounds inefficient.. Is it too wild to create JNI bindings for `BTreeMap` itself?

twitu:
The JNIX bindings can be put behind a feature flag and they will only be compiled when the feature flag is set on. Separating the bindings from the logic may not be very helpful it is closely tied to the logic.

There's no good way to wrap a generic Result type in a way that's accessible to Java. Option is doable but then it will lose out on a lot of useful information. I guess in this case the best approach is to return Java Exceptions and the let the library user figure out how to handle it and it some rare cases return a default value where it makes sense.

I looked in implementation for wrapping container types, especially using `jnix`. It converts container types to Java in a way that's very similar to the one we've implemented for `LinkedHashMap` basically initializing the java object and inserting values into it. Very bad from an efficiency standpoint.

A more efficient way will be to keep the BTreeMap on the Rust side and define a specific set of operations on it through the jni bindings. I guess the most basic operation will be to `get` the value for a key. Perhaps a method to iterate over all the values. Do you think any other operations will be useful to library users?

fs-storage/src/FileStorage.h
```diff
@@ -0,0 +1,69 @@
+/* DO NOT EDIT THIS FILE - it is machine generated */
```
kirillt, line 1: Do we really need this file? We had bindings without C code before.

fs-storage/src/file_storage.rs
```diff
@@ -356,3 +356,170 @@ mod tests {
         assert_eq!(file_storage_1.as_ref().get("key3"), Some(&9));
     }
 }
+
+#[cfg(feature = "jni-bindings")]
+pub mod jni_bindings {
```
tareknaser, line 361: Could we have this module in a separate file?

fs-storage/tests/Makefile
```diff
@@ -0,0 +1,23 @@
+CARGO_CMD = cargo
+RUST_BUILD_FLAGS = --release --features jni-bindings
+JAR_FILE := junit.jar
+URL := https://repo1.maven.org/maven2/org/junit/platform/junit-platform-console-standalone/1.9.3/junit-platform-console-standalone-1.9.3.jar
+
+.PHONY: test
+test: build-rust download-test-lib java-test
```
tareknaser, line 7: Can we run this in the CI as well?

fs-storage/src/utils.rs
```diff
@@ -15,15 +15,8 @@ use std::path::Path;
...
 where
-    K: Ord
-        + Clone
-        + serde::Serialize
-        + serde::de::DeserializeOwned
-        + std::str::FromStr,
-    V: Clone
-        + serde::Serialize
-        + serde::de::DeserializeOwned
-        + std::str::FromStr,
```
tareknaser, line 26: Why do we need to remove `Clone`?

fs-storage/tests/Makefile
tareknaser, line 1: Why is the `Makefile` located in `fs-storage/tests/Makefile`? Shouldn't it be in `fs-storage/Makefile` instead?

fs-storage/src/file_storage.rs
```diff
@@ -356,3 +356,170 @@ mod tests {
...
+
+    #[no_mangle]
+    pub extern "system" fn Java_FileStorage_writeFS(
+        _env: &mut JNIEnv,
+        _class: JClass,
+        file_storage_ptr: jlong,
+    ) {
+        match FileStorage::from_jlong(file_storage_ptr).write_fs() {
+            Ok(_) => {}
+            Err(_) => {} // handle error here
```
tareknaser, line 507: Why are we overlooking errors here?

tareknaser (commented):
Thanks. 
I'm not sure I can review the Java part, but I left some general comments.

fs-storage/tests/FileStorageTest.java
```diff
@@ -0,0 +1,87 @@
...
+    @Test
+    public void testFileStorageNeedsSyncing() {
+        String label = "test";
+        Path storagePath = tempDir.resolve("test.txt");    
+        FileStorage fileStorage = new FileStorage(label, storagePath.toString());
+        assertFalse(fileStorage.needsSyncing());
+
+        fileStorage.set("key1", "value1");
+        // FAIL: don't why it is still false
+        // assertTrue(fileStorage.needsSyncing());
```
kirillt, line 59: If it's a bug in the `fs-storage` we should register new issue for this.

fs-storage/src/utils.rs
kirillt: @tareknaser isn't it good if we can loosen requirements for `V`?

fs-storage/src/file_storage.rs
```diff
@@ -356,3 +356,170 @@ mod tests {
...
+        _class: JClass,
+        file_storage_ptr: jlong,
+    ) {
+        let file_storage = unsafe {
+            Box::from_raw(file_storage_ptr as *mut FileStorage<String, String>)
+        };
+        file_storage.erase().unwrap();
+    }
+
+    // TODO: Implement mergeFrom
```
kirillt, line 524: Should be implemented in this PR, or new issued should be registered.

fs-storage/src/file_storage.rs
```diff
@@ -356,3 +356,170 @@ mod tests {
...
+        file_storage_ptr: jlong,
+    ) -> jobject {
+        let data: BTreeMap<String, String> =
+            FileStorage::from_jlong(file_storage_ptr)
+                .read_fs()
+                .unwrap();
+
+        // Create a new LinkedHashMap object
+        let linked_hash_map_class =
+            env.find_class("java/util/LinkedHashMap").unwrap();
```
kirillt, line 464: Am I right that this is temporary until we implement bindings for `BTreeMap`?

fs-storage/README.md
kirillt, line 1: @tareknaser If you could validate updates in README, that would be great. Ideally, all Rust developers in our project should be able to work with bindings.

fs-storage/src/file_storage.rs
Pushkarm029: @twitu

fs-storage/tests/FileStorageTest.java
Pushkarm029: ok, creating a new issue

Pushkarm029:
CI is failing due to Java warnings. It is running perfectly locally.

tareknaser:
The logs from the failing CI show that there is probably an error not just warnings 

```sh
cd tests && \
if [ ! -f "junit.jar" ]; then \
	curl -o "junit.jar" "https://repo1.maven.org/maven2/org/junit/platform/junit-platform-console-standalone/1.9.3/junit-platform-console-standalone-1.9.3.jar"; \
else \
	echo "junit.jar already exists."; \
fi
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100 2552k  100 2552k    0     0  9533k      0 --:--:-- --:--:-- --:--:-- 9525k
(cd ../target/debug && \
export LD_LIBRARY_PATH=$PWD && \
cd ../../fs-storage/tests && \
javac -Xlint:none -d out FileStorage.java && \
javac -Xlint:none -d out -cp out:junit.jar FileStorageTest.java && \
java -jar junit.jar --class-path out --scan-class-path)
Note: FileStorageTest.java uses unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.
thread '<unnamed>' panicked at /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/jni-0.21.1/src/wrapper/jnienv.rs:520:9:
NullDeref("*JNIEnv")
fatal runtime error: failed to initiate panic, error 5
Aborted (core dumped)
make: *** [Makefile:24: java-test] Error 134
```

Pushkarm029:
locally i am getting this : (im on debian)
```bash
pushkarm029@thinkpad:~/ark-rust(jni-bindings○) » cd fs-storage
pushkarm029@thinkpad:~/ark-rust/fs-storage(jni-bindings○) » make test
cargo build --release --features jni-bindings
    Finished `release` profile [optimized] target(s) in 0.24s
cd tests && \
if [ ! -f "junit.jar" ]; then \
	curl -o "junit.jar" "https://repo1.maven.org/maven2/org/junit/platform/junit-platform-console-standalone/1.9.3/junit-platform-console-standalone-1.9.3.jar"; \
else \
	echo "junit.jar already exists."; \
fi
junit.jar already exists.
(cd ../target/debug && \
export LD_LIBRARY_PATH=$PWD && \
cd ../../fs-storage/tests && \
javac -Xlint:none -d out FileStorage.java && \
javac -Xlint:none -d out -cp out:junit.jar FileStorageTest.java && \
java -jar junit.jar --class-path out --scan-class-path)
Note: FileStorageTest.java uses unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.

Thanks for using JUnit! Support its development at https://junit.org/sponsoring

╷
├─ JUnit Jupiter ✔
│  └─ FileStorageTest ✔
│     ├─ testFileStorageNeedsSyncing() ✔
│     ├─ testFileStorageMonoidCombine() ✔
│     ├─ testFileStorageAutoDelete() ✔
│     ├─ testFileStorageMainScenario() ✔
│     └─ testFileStorageWriteRead() ✔
├─ JUnit Vintage ✔
└─ JUnit Platform Suite ✔

Test run finished after 186 ms
[         4 containers found      ]
[         0 containers skipped    ]
[         4 containers started    ]
[         0 containers aborted    ]
[         4 containers successful ]
[         0 containers failed     ]
[         5 tests found           ]
[         0 tests skipped         ]
[         5 tests started         ]
[         0 tests aborted         ]
[         5 tests successful      ]
[         0 tests failed          ]

```

Pushkarm029:
Yup, there might be some internal issue, but I will try to fix it.

fs-storage/tests/FileStorageTest.java
Pushkarm029: #63 ig this will fix this.

Pushkarm029:
Finally, Java Linux CI Fixed.

Working on others.....

Pushkarm029:
MacOS fixed

fs-storage/src/jni_file_storage.rs
```diff
@@ -0,0 +1,185 @@
...
+        match FileStorage::from_jlong(file_storage_ptr).needs_syncing() {
+            Ok(updated) => updated as jboolean,
+            Err(err) => {
+                env.fatal_error(err.to_string());
+            }
+        }
+    }
+
+    #[no_mangle]
+    pub extern "system" fn Java_FileStorage_readFS(
```
kirillt, line 96: Resorting to Java's `LinkedHashMap` could be fine for this prototype, but I think we should try the approach with writing our own bindings to `BTreeMap`.

fs-storage/tests/FileStorage.java
```diff
@@ -0,0 +1,69 @@
+public class FileStorage {
```
kirillt, line 1: Could you create the counter-part PR in [ark-android](https://github.com/ARK-Builders/ark-android) which uses this branch as backend for [the storage demo](https://github.com/ARK-Builders/ark-android/pull/97)?

fs-storage/tests/FileStorage.java
```diff
@@ -0,0 +1,69 @@
...
+
+    public void erase() {
+        erase(this.fileStoragePtr);
+    }
+
+    public void merge(FileStorage other) {
+        merge(this.fileStoragePtr, other.fileStoragePtr);
+    }
+
+    public static void main(String[] args) {
```
kirillt, line 56: Unit tests should be enough, no need to create `main` entrypoint.

fs-storage/tests/FileStorageTest.java
```diff
@@ -0,0 +1,123 @@
...
+    @Test
+    public void testFileStorageNeedsSyncing() {
+        String label = "test";
+        Path storagePath = tempDir.resolve("test.txt");    
+        FileStorage fileStorage = new FileStorage(label, storagePath.toString());
+        fileStorage.writeFS();
+        assertFalse(fileStorage.needsSyncing());
+        fileStorage.set("key1", "value1");
+        // // FAIL: don't why it is still false
+        // assertTrue(fileStorage.needsSyncing());
```
kirillt, line 61: Should be fixed before merging.

kirillt:
I believe we should merge this PR first:
- #63 

Also, regarding TODO points from the PR description:

- [ ] Store the `jlong` in the class instance itself rather than returning it
> It will reduce the amount of Java/Kotlin wrapper code that needs to be written

- [ ] Handle Results better
> Should they return a default value or should they thrown an exception? open to discussion

I think, functional way with `Result` is better especially because Kotlin adopts features from functional paradigm and has native `Result` type. If it's difficult to pass some error type, we can just use `String` message. 

- [ ] Refactor `read_fs`
> It returns the key value pairs as a BTreeMap, however jni by default does not know how to convert BTreeMap to Java's LinkedHashMap. Do we want to support the `read_fs` operation for JNI bindings? This means we'll have to implement a custom conversion logic from BTreeMap to LinkedHashMap

We can start with the conversion and integrate with [the storage sample](https://github.com/ARK-Builders/ark-android/pull/97) first, but would be great to write bindings to `BTreeMap` directly. We can do it in separate PR though.

twitu (commented):
@Pushkarm029 in your last commit returning `jstring` from all functions is not a proper solution to handling errors.

A Rust `Result<T>` cannot be returned directly across jni bindings because bindings don't allow passing generics. There is no good way to communicate if the Result contains a bool, or integer or string or something else.

There are two possible options to consider -

* Return exceptions - throw exception using `throw` function in JNIEnv. Perhaps mplement a conversion from ArkStorageError to jni::error::Exception if it's not ergonomic by default.

* Implement a conversion to Kotlin Result type[^1] by creating a native java object similar to how a BTreeMap is transformed into a LinkedHashMap.

Note: that the error message in both cases will be a string version of the ArkStorageError


[^1]: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-result/

twitu:
>     * [ ]  Store the `jlong` in the class instance itself rather than returning it

The main purpose of this point was to avoid the writing the Java wrapper code like the one implemented in `FileStorage.java`. This java code wraps around the static methods exposed by the rust jni bindings. I wanted to avoid this by hiding the state of the FileStorage and directly exposing the a FileStorage Java object.

However, that does not seem to be possible based on my exploration. There is no way to create Java class instances with hidden state[^1] on the Rust side that can then be transferred over through JNI bindings. Even looking through the documentation of jni and jnix there is no way to define a Java "class" on the Rust side. It only provides functionality to convert between Rust and Java data types and export static methods through JNI bindings.

As such, I believe this PR is ready to merge. The syncing logic PR can adopt these changes as it might take more time to identify and fix the timestamp bug in it.

[^1]: https://stackoverflow.com/questions/25025956/create-anonymous-java-class-in-c-via-jni

fs-storage/tests/FileStorageTest.java
```diff
@@ -55,13 +57,16 @@ public void testFileStorageNeedsSyncing() {
...
-        assertFalse(fileStorage.needsSyncing());
+        String result = fileStorage.needsSyncing();
+        assertEquals("false", result);
         fileStorage.set("key1", "value1");
         // // FAIL: don't why it is still false
         // assertTrue(fileStorage.needsSyncing());
 
-        fileStorage.writeFS();
-        assertFalse(fileStorage.needsSyncing());
+        String err = fileStorage.writeFS();
```
kirillt: Can we uncomment the assertion now?
```
        fileStorage.set("key1", "value1");
        // // FAIL: don't why it is still false
        // assertTrue(fileStorage.needsSyncing());
```

fs-storage/tests/FileStorageTest.java
kirillt: @Pushkarm029

fs-storage/README.md
kirillt: @tareknaser

fs-storage/tests/FileStorage.java
kirillt: @twitu @Pushkarm029

kirillt:
@twitu Could you also create tracking issue for functional error handling (`Result`)?

And we need the sample wired to this branch, that would be the best demo.

twitu (commented):
@Pushkarm029 the new syncing logic has introduced a new struct for `SyncStatus` that must be exposed through JNI along with a couple of functions added to the `BaseStorage` trait. Some of the existing tests and functions need to be updated.

```rust
    /// Get [`SyncStatus`] of the storage
    fn sync_status(&self) -> Result<SyncStatus>;

    /// Sync the in-memory storage with the storage on disk
    fn sync(&mut self) -> Result<()>;
```

One way to expose `SyncStatus` is using the `jnix` crate which will automatically derive the conversion logic to and from java objects. Note that jnix should be feature flagged just like jni.

Kotlin style Results is scoped in a separate issue #74

fs-storage/tests/FileStorageTest.java
Pushkarm029: yes

fs-storage/tests/FileStorage.java
twitu: As we discussed this will be done in a separate PR, once the ark-rust repo is refactored to move all java related code into a single java project.

Pushkarm029:
done; pls review @twitu @tareknaser

fs-storage/src/base_storage.rs
```diff
@@ -1,7 +1,9 @@
 use data_error::Result;
+use jnix_macros::{FromJava, IntoJava};
 use std::collections::BTreeMap;
 
-#[derive(Debug, PartialEq, PartialOrd, Ord, Eq, Clone)]
+#[derive(Debug, PartialEq, PartialOrd, Ord, Eq, Clone, FromJava, IntoJava)]
+#[jnix(class_name = "SyncStatus")]
```
twitu, line 6: Use a conditional derive for `FromJava` and `IntoJava`. This is so that Java bindings are completely featured flagged and it's possible to turn it off if a user only wants to import the Rust core.

```
#[cfg_attr(feature = "<feature>", derive(Stuff))]
```

Similarly for the `jnix_macros` import.

fs-storage/src/base_storage.rs
```diff
@@ -1,7 +1,9 @@
 use data_error::Result;
+use jnix_macros::{FromJava, IntoJava};
```
twitu, line 2: Why is `jnix_macros` preferable over `jnix` which also exports `jnix_macros` under a feature flag?

fs-storage/tests/SyncStatus.java
```diff
@@ -0,0 +1,11 @@
+/// Represents the synchronization status of the storage.
+public enum SyncStatus {
```
twitu, line 2: It's unfortunate that we cannot avoid duplicating the SyncStatus definition. I'd hoped the `jnix_macro` should would create a class definition as part of the exported package.

twitu (commented):
A minor nit related to feature flagging.

fs-storage/src/base_storage.rs
Pushkarm029: done

twitu (commented):
MERGE! :rocket:

fs-storage/Cargo.toml
```diff
@@ -15,11 +15,16 @@ name = "cli"
 log = { version = "0.4.17", features = ["release_max_level_off"] }
 serde_json = "1.0.82"
 serde = { version = "1.0.138", features = ["derive"] }
-
+jni = { version = "0.21.1", optional = true }
+jnix = { version = "0.5.1", features = ["derive"] }
```
tareknaser, line 19: I guess `jnix` should be optional as well

.gitignore
tareknaser, line 1: We should also git ignore `*.h` files

fs-storage/README.md
```diff
@@ -44,3 +44,15 @@ key2: value2
 ```bash
 cargo run --example cli read /tmp/z
 ```
+
+## Java Wrapper
+```java
+javac -h . FileStorage.java
+javac FileStorage.java  
+LD_LIBRARY_PATH=<project-root-path>/target/debug && java FileStorage.java
```
tareknaser, line 52: The project root path is always `../` if we run from `fs-storage/`

fs-storage/tests/SyncStatus.java
tareknaser, line 2: We should a note in the rust code of `SyncStatus` that any change should be applied here as well

fs-storage/Makefile
```diff
@@ -0,0 +1,52 @@
...
+.PHONY: test-windows
+test-windows: build-rust download-test-lib java-test-windows
+
+.PHONY: build-rust
+build-rust:
+	$(CARGO_CMD) build
+
+.PHONY: download-test-lib
+download-test-lib: 
+	cd tests && \
```
tareknaser, line 20:
```suggestion
	@cd tests && \
```

So it doesn't print the command to the terminal each time

tareknaser (approved):
I have left some comments, but I think we should merge. These comments can be addressed in #75.

Pushkarm029:
Thanks @tareknaser for the review, i completed all these suggestions in #76

PR #63, opened 2024-05-20
New syncing logic

### LINKED ISSUE #37: `fs-storage`: Millisecond timestamps and proper rounding
Context: https://github.com/ARK-Builders/ark-rust/pull/23#discussion_r1572807479

We might need more granular timestamps than just seconds. Let's consider such a scenario:
1. We scan the filesystem at the moment `HH:MM:ss:xxx`, where `xxx` is milliseconds part.
2. A file `cat.jpg` is modified at the moment `HH:MM:ss:yyy`.

If we truncate `xxx` and `yyy` then on next scan we'll skip `cat.jpg` and we miss the update. We has performed the scan in the moment `HH:MM:ss` and the file has been modified at `HH:MM:ss`, so this doesn't look like an update.

As a workaround, it seems possible to round `HH:MM:ss:xxx` always to the lowest side when we memoize the timestamp of latest scan, and to round `HH:MM:ss:yyy` always to the highest side when we retrieve the timestamp of file modification. This would result in excessive updates for the files which were updated in the same second when scan happened, but it should be better than losing an update.

With this logic implemented, we control amount of excessive scans by altering the timestamp granularity. What is the smallest division of time that makes sense for file systems? Nanoseconds seem to be too low, but what about milliseconds?

twitu:
Closes #37


Pushkarm029 (approved):

kirillt:
I think we have also discussed "smart rounding":
> round HH:MM:ss:xxx always to the lowest side when we memoize the timestamp of latest scan
> round HH:MM:ss:yyy always to the highest side when we retrieve the timestamp of file modification

twitu:
SystemTime records time at a nanosecond level depending on the operating system. The `as_millis` function implements the standard logic for rounding the timestamp up to milliseconds. The default implementation should be sufficient for our purpose.

```rust
    pub const fn as_millis(&self) -> u128 {
        self.secs as u128 * MILLIS_PER_SEC as u128 + (self.nanos.0 / NANOS_PER_MILLI) as u128
    }
```

twitu:
If we want to reduce the possibility of a race condition, we can use nano second resolution without any loss in terms of performance and increase in complexity. We are already is storing the SystemTime which is represented as a `u128`.

```rust
    pub const fn as_nanos(&self) -> u128 {
        self.secs as u128 * NANOS_PER_SEC as u128 + self.nanos.0 as u128
    }
```

kirillt:
@twitu yes, nanos are possible and are provided by the filesystem API, but iirc they are not real nanos and there are zeros at the end.. But why you don't want to adopt different rounding logic for reading timestamp? I think this would solve the issue for any precision.

twitu:
I just noticed that because of the recent changes in the monoids PR. The semantics for the syncing logic has changed. We're now checking for timestamp equality, so rounding logic won't work. We can only increase the resolution of the timestamp.

```rust
    /// Compare the timestamp of the storage file
    /// with the timestamp of the in-memory storage update
    /// to determine if either of the two requires syncing.
    fn needs_syncing(&self) -> Result<bool> {
        match fs::metadata(&self.path) {
            Ok(metadata) => {
                let get_duration_since_epoch = |time: SystemTime| {
                    time.duration_since(SystemTime::UNIX_EPOCH)
                        .unwrap()
                        .as_secs()
                };

                let fs_modified =
                    get_duration_since_epoch(metadata.modified()?);
                let self_modified = get_duration_since_epoch(self.modified);

                Ok(fs_modified != self_modified)
```

kirillt:
> I just noticed that because of the recent changes in the monoids PR. The semantics for the syncing logic has changed. We're now checking for timestamp equality, so rounding logic won't work. We can only increase the resolution of the timestamp.

Agree, this changes everything...

Ready to merge as soon as build is fixed.

twitu:
A completely revamped syncing logic. This affects `read_fs`, `write_fs`, `new` FileStorage and some minor changes to the `BaseStorage` api.

Needs a solid review.

fs-storage/src/file_storage.rs
```diff
@@ -176,9 +157,77 @@ where
...
+        + serde::Serialize
+        + serde::de::DeserializeOwned
+        + std::str::FromStr
+        + Monoid<V>,
+{
+    /// Set a key-value pair in the storage
+    fn set(&mut self, key: K, value: V) {
+        self.data.entries.insert(key, value);
+        self.modified = std::time::SystemTime::now();
+    }
```
tareknaser, line 182: Why not write to the file system whenever a user sets a value?

fs-storage/src/file_storage.rs
```diff
@@ -176,9 +157,77 @@ where
...
+
+    /// Remove a key-value pair from the storage given a key
+    fn remove(&mut self, id: &K) -> Result<()> {
+        self.data.entries.remove(id).ok_or_else(|| {
+            ArklibError::Storage(self.label.clone(), "Key not found".to_owned())
+        })?;
+        self.modified = std::time::SystemTime::now();
+        Ok(())
+    }
```
tareknaser, line 191: Here too. 

If a user removes a value from `FileStorage` instance, I think we should update file system to reflect the changes. 
What do you think?

fs-storage/src/base_storage.rs
```diff
@@ -1,27 +1,37 @@
...
+    /// Load key-value mapping from the file system.
+    UpSync,
+    /// Write key-value mapping to the file system.
+    DownSync,
+    /// Changes from both in-memory and file system need to be merged before syncing.
+    FullSync,
+}
+
+/// A trait for a key-value mapping that is persisted to the file system.
 pub trait BaseStorage<K, V>: AsRef<BTreeMap<K, V>> {
```
tareknaser, line 18: I think we should actually start here. The `BaseStorage` trait's methods should have explicit documentation comments explaining their functionality. Questions such as whether to write to the file system when setting/removing a value, when to sync, and when to call `merge_from` should be clearly stated in the trait itself, as these methods work together. 

The implementation in `FileStorage` should only contain the logic defined in the trait.

fs-storage/src/file_storage.rs
```diff
@@ -176,9 +157,77 @@ where
...
+            }
+        } else {
+            // file updated since last write
+            if file_updated > self.written_to_disk {
+                // only file updated
+                return Ok(SyncStatus::UpSync);
+            } else {
+                // no change
+                return Ok(SyncStatus::NoSync);
+            }
```
tareknaser, line 217: I'm sorry, but could you explain the logic here a bit more? I'm not sure I get it.

fs-storage/src/file_storage.rs
```diff
@@ -62,75 +72,43 @@ where
...
-    pub fn new(label: String, path: &Path) -> Self {
-        Self {
+    pub fn new(label: String, path: &Path) -> Result<Self> {
+        let time = SystemTime::now();
+        let mut storage = Self {
             label,
             path: PathBuf::from(path),
-            modified: SystemTime::now(),
+            modified: time,
+            written_to_disk: time,
```
tareknaser, line 81: When do these 2 fields have different values?

fs-storage/src/file_storage.rs
```diff
@@ -299,34 +355,35 @@ mod tests {
...
         let mut file_storage =
-            FileStorage::new("TestStorage".to_string(), &storage_path);
+            FileStorage::new("TestStorage".to_string(), &storage_path).unwrap();
         file_storage.write_fs().unwrap();
-        assert_eq!(file_storage.needs_syncing().unwrap(), false);
+        assert_eq!(file_storage.needs_syncing().unwrap(), SyncStatus::NoSync);
         std::thread::sleep(std::time::Duration::from_secs(1));
         file_storage.set("key1".to_string(), "value1".to_string());
-        assert_eq!(file_storage.needs_syncing().unwrap(), true);
+        assert_eq!(file_storage.needs_syncing().unwrap(), SyncStatus::DownSync);
```
tareknaser, line 363: Thank you for adding the tests. 

Could you include additional tests for the `SyncStatus::UpSync` and `SyncStatus::FullSync` cases?

fs-storage/src/file_storage.rs
tareknaser, line 1: The `test_file_storage_is_storage_updated` test is failing on macOS. Could you take a look at it? 

Here is the output:

```sh
---- file_storage::tests::test_file_storage_is_storage_updated stdout ----
thread 'file_storage::tests::test_file_storage_is_storage_updated' panicked at fs-storage/src/file_storage.rs:360:9:
error: test failed, to rerun pass `-p fs-storage --lib`
  left: UpSync
 right: NoSync
```

tareknaser (commented):
Thank you for putting this together. I've left some comments. 

There are also a few clippy warnings causing the CI to fail. I've reviewed them, and they're straightforward to address. 
You can run this command to see them
```sh
cargo clippy --workspace --bins -- -D warnings
```

fs-storage/src/file_storage.rs
kirillt: We've discussed this in the other PR. The idea is to allow the user to control when exactly the write is happening. Performance reasons mostly. Storage values can be updated often, but writing to disk is a slow operation. Also, we run coroutines in Kotlin to update storages. It's difficult to do sequentially because it's IO and we don't want to freeze UI thread. So, if we create lots of coroutines they start race with each other. We have atomic versions exactly for this reason, but it's also nice to just reduce the load.

fs-storage/src/file_storage.rs
kirillt: Explained the decision in comment to `fn set`.

fs-storage/src/file_storage.rs
tareknaser: Thanks for clarifying. I think I had forgotten about this.

fs-storage/src/base_storage.rs
```diff
@@ -1,27 +1,37 @@
...
+pub enum SyncStatus {
+    /// No synchronization needed.
+    NoSync,
+    /// Load key-value mapping from the file system.
+    UpSync,
+    /// Write key-value mapping to the file system.
+    DownSync,
+    /// Changes from both in-memory and file system need to be merged before syncing.
+    FullSync,
+}
```
kirillt, line 15: Could you describe the motivation behind this separation? My guess is that `DownSync` is what should be used with cache storages, and any other storages seem to require `FullSync`? But when do we need `UpSync` or `NoSync`?

kirillt, line 15: Probably I would reword names of these values like "InSync", "Ours", "Theirs", "Both" to accent on which side should be updated. This would also make the enum potentially re-usable with other kinds of backend (filesystem is only starting point, later we might use Iroh directly).

fs-storage/src/file_storage.rs
```diff
@@ -100,37 +109,29 @@ where
...
+
+        match (
+            self.modified > self.written_to_disk,
+            self.written_to_disk == file_updated,
+            file_updated > self.written_to_disk,
+        ) {
+            (true, true, _) => Ok(SyncStatus::DownSync),
+            (true, false, _) => Ok(SyncStatus::FullSync),
+            (_, _, true) => Ok(SyncStatus::UpSync),
+            _ => Ok(SyncStatus::NoSync),
```
kirillt: This looks cool 👍

kirillt: This looks cool 👍

fs-storage/src/file_storage.rs
kirillt: Actually, previous `match` expression looked more readable to me.

@tareknaser Ishan introduced one more timestamp, so now we track times of both internal modification and writing to the filesystem. We have third timestamp in the filesystem itself. These 3 timestamps are enough to figure out which side updated, or if both sides been updated.

fs-storage/src/base_storage.rs
twitu:
```
    /// No synchronization needed.
    InSync,
    /// In-memory key-value mapping is stale.
    MappingStale,
    /// External file system storage is stale.
    StorageStale,
    /// In-memory key-value mapping and external file system storage diverge.
    Diverge,
```

How about this? Ours and theirs can become confusing. Since this dealing with FileStorage, mapping and storage makes it clear what we're refering to.

fs-storage/src/base_storage.rs
twitu: I've updated the comments. I've made it clear that the BaseStorage trait only updates the mapping and it is the responsibilty of the implementor to decide when to interact with the file system.

fs-storage/src/file_storage.rs
twitu: When a value is set only `modified` is updated to current time. So it becomes `modified > written_to_disk` which shows that mapping is ahead of file.

When `write_fs` or `read_fs` is called it becomes `modified = written_to_disk` to reflect that mapping and file are in sync.

fs-storage/src/file_storage.rs
```diff
@@ -62,75 +72,43 @@ where
...
-            }
-            Err(e) => {
-                Err(ArklibError::Storage(self.label.clone(), e.to_string()))
-            }
+    pub fn sync(&mut self) -> Result<()> {
+        match self.needs_syncing()? {
+            SyncStatus::NoSync => Ok(()),
+            SyncStatus::DownSync => self.read_fs().map(|_| ()),
+            SyncStatus::UpSync => self.write_fs(),
+            SyncStatus::FullSync => self.full_sync(),
```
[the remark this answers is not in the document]
twitu: I've added the log statement to the inner function - `sync_status` since it might be called in other places as well.

fs-storage/src/file_storage.rs
```diff
@@ -62,75 +72,43 @@ where
...
+            written_to_disk: time,
             data: FileStorageData {
                 version: STORAGE_VERSION,
                 entries: BTreeMap::new(),
             },
+        };
+
+        if Path::exists(path) {
+            let _ = storage.read_fs();
         }
```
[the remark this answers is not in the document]
twitu: Fixed.

fs-storage/src/file_storage.rs
twitu: Yes there is some race condition happening there. It's good that the tests have caught it. I'll fix the syncing logic.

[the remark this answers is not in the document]
twitu: I've added `sync` to `BaseStorage` trait and added doc comments. I've inlined full_sync since it's reusing BaseStorage methods and FileStorage specific.

fs-storage/src/base_storage.rs
kirillt: This looks good. Ours/theirs is kinda conventional but I agree that it can be confusing.

fs-storage/src/file_storage.rs
kirillt: I think it's worth to document these invariants as comments to `modified` and `written_to_disk` variables.

fs-storage/src/file_storage.rs
```diff
@@ -62,75 +72,28 @@ where
         + Monoid<V>,
 {
     /// Create a new file storage with a diagnostic label and file path
-    pub fn new(label: String, path: &Path) -> Self {
-        Self {
+    pub fn new(label: String, path: &Path) -> Result<Self> {
```
kirillt, line 75: Also worth to mention in the comment that it's initialized with the data stored by the path, if the path exists.

fs-storage/src/file_storage.rs
```diff
@@ -176,12 +142,100 @@ where
...
+        + serde::Serialize
+        + serde::de::DeserializeOwned
+        + std::str::FromStr,
+    V: Clone
+        + serde::Serialize
+        + serde::de::DeserializeOwned
+        + std::str::FromStr
+        + Monoid<V>,
+{
+    /// Set a key-value pair in the storage
```
kirillt, line 163:
```suggestion
    /// Set a key-value pair in the internal mapping
```

fs-storage/src/file_storage.rs
```diff
@@ -176,12 +142,100 @@ where
...
+    fn set(&mut self, key: K, value: V) {
+        self.data.entries.insert(key, value);
+        self.modified = std::time::SystemTime::now();
+    }
+
+    /// Remove a key-value pair from the storage given a key
```
kirillt, line 169:
```suggestion
    /// Remove an entry from the internal mapping given a key
```

fs-storage/src/file_storage.rs
```diff
@@ -62,75 +72,28 @@ where
         + Monoid<V>,
 {
     /// Create a new file storage with a diagnostic label and file path
```
kirillt, line 74:
```suggestion
    /// Create a new file storage with a diagnostic label and file path
    /// The storage will be initialized using the disk data, if the path exists
```

fs-storage/src/file_storage.rs
```diff
@@ -176,12 +142,100 @@ where
...
+            SyncStatus::MappingStale => self.read_fs().map(|_| ()),
+            SyncStatus::StorageStale => self.write_fs(),
+            SyncStatus::Diverge => {
+                let data = self.load_fs_data()?;
+                self.merge_from(&data)?;
+                self.write_fs()?;
+                Ok(())
+            }
+        }
+    }
```
kirillt, line 224: 👍

fs-storage/src/file_storage.rs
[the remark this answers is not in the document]
tareknaser: Nice. Thank you

fs-storage/src/file_storage.rs
```diff
@@ -62,75 +78,29 @@ where
         + Monoid<V>,
 {
     /// Create a new file storage with a diagnostic label and file path
-    pub fn new(label: String, path: &Path) -> Self {
-        Self {
+    /// The storage will be initialized using the disk data, if the path exists
```
tareknaser, line 81:
```suggestion
    /// The storage will be initialized using the disk data, if the path exists
    ///
    /// Note: if the file storage already exists, the data will be read from the file
    /// without overwriting it.
```

fs-storage/src/file_storage.rs
```diff
@@ -176,12 +149,100 @@ where
...
+    /// with the timestamp of the in-memory storage and the last written
+    /// to time to determine if either of the two requires syncing.
+    fn sync_status(&self) -> Result<SyncStatus> {
+        let file_updated = fs::metadata(&self.path)?.modified()?;
+
+        // mapping updated
+        let status = if self.modified > self.written_to_disk {
+            // file updated since last write
+            if file_updated > self.written_to_disk {
+                // both updated
```
tareknaser, line 195: We are only here if the user manually modifies the storage file on disk, correct?

fs-storage/src/file_storage.rs
```diff
@@ -292,41 +361,73 @@ mod tests {
         assert_eq!(storage_path.exists(), false);
     }
 
+    #[test]
+    fn test_file_timestamp_bug() {
```
tareknaser, line 365: Could we remove this test case now that we've fixed the bug?

tareknaser (commented):
Thanks for all your work on this! I think we're almost ready to merge. 

Could you also rebase it on top of main?

fs-storage/src/file_storage.rs
twitu: Externally modified yes. Either by the user or suppose by a mirror storage (like in the test).

twitu:
Another weird timestamp related bug. It seems `write_fs` doesn't even change the modified timestamp in the file metadata.

```
before_write: SystemTime { tv_sec: 1718114429, tv_nsec: 471980190 }, after_write: SystemTime { tv_sec: 1718114429, tv_nsec: 471980190 }
thread 'file_storage::tests::test_file_storage_is_storage_updated' panicked at fs-storage/src/file_storage.rs:406:9:
```

tareknaser:
> Another weird timestamp related bug. It seems `write_fs` doesn't even change the modified timestamp in the file metadata.

Could you write a minimal test to reproduce it? 
locally for me, all tests are passing at 4a13396

twitu (commented):
Added an MRE in the failing `test_file_metadata_timestamp_updated` test

tareknaser:
> Added an MRE in the failing test_file_metadata_timestamp_updated test

It looks like this issue might be related to the file system's timestamp granularity. From what I've found, it seems to depend on the file system being used.

For Linux:
- `EXT3` supports timestamps with a precision up to one second.
- `EXT4` can support timestamps with a precision up to nanoseconds, but [not always](https://stackoverflow.com/a/76581457)
> I am not entirely sure about these numbers. I couldn't find a definitive resource to list them.

I've pushed a commit that adds a one-second delay after each file write to fix the tests. This seems to be enough to fix the tests. I also reviewed the Rust file system documentation and added a call to [`sync_all()`](https://doc.rust-lang.org/std/fs/struct.File.html#method.sync_all).

I’m wondering if users might run into similar issues with `fs-storage`. It might be worth adding a note about this in the `FileStorage`’s doc comment.

twitu:
It's unfortunate that the OS does not update the metadata fast enough for it to be useful for us. However, we can perhaps work around it by setting the modified time ourself. With this logic we can set the modified timestamp consistently and the tests pass without needing a sleep.

```rust
        fs::create_dir_all(parent_dir)?;
        let mut file = File::create(&self.path)?;
        file.write_all(serde_json::to_string_pretty(&self.data)?.as_bytes())?;
        file.flush()?;

        let new_timestamp = SystemTime::now();
        file.set_modified(new_timestamp)?;
        file.sync_all()?;
```

fs-storage/src/file_storage.rs
```diff
@@ -29,9 +29,16 @@ pub struct FileStorage<K, V>
 where
     K: Ord,
 {
+    /// Label for logging
     label: String,
+    /// Path to underlying file where data is persisted
```
tareknaser, line 34:
```suggestion
    /// Path to the underlying file where data is persisted
```

fs-storage/src/file_storage.rs
```diff
@@ -190,16 +251,16 @@ where
...
-        if new_timestamp == self.modified {
-            return Err("Timestamp has not been updated".into());
-        }
+        let mut file = File::create(&self.path)?;
+        file.write_all(serde_json::to_string_pretty(&self.data)?.as_bytes())?;
+        file.flush()?;
+
+        let new_timestamp = SystemTime::now();
+        file.set_modified(new_timestamp)?;
+        file.sync_all()?;
```
tareknaser, line 260: Thanks. This is a cool solution!

fs-storage/src/file_storage.rs
```diff
@@ -293,40 +360,79 @@ mod tests {
...
         std::thread::sleep(std::time::Duration::from_secs(1));
+
         file_storage.set("key1".to_string(), "value1".to_string());
-        assert_eq!(file_storage.needs_syncing().unwrap(), true);
+        let before_write = fs::metadata(&storage_path)
+            .unwrap()
+            .modified()
+            .unwrap();
+        file_storage.write_fs().unwrap();
+        std::thread::sleep(std::time::Duration::from_secs(1));
```
tareknaser, line 379: Since we now have a clever solution, can we remove all the delays from the tests?

tareknaser (approved):
Thanks for making all the requested changes. I've added a few minor comments. 

Once you rebase, we should be ready to merge.

fs-storage/src/file_storage.rs
```diff
@@ -176,12 +152,97 @@ where
...
         self.modified = fs::metadata(&self.path)?.modified()?;
+        self.written_to_disk = self.modified;
+        self.data = data;
 
-        Ok(data.entries)
+        Ok(&self.data.entries)
     }
 
-    /// Write the data to the storage file
+    /// Write the data to file
```
tareknaser, line 245: We should also mention the trick of manually modifying the timestamp and why we need it as a note in the doc comment of this method

kirillt:
Good job @twitu @Pushkarm029 @tareknaser, this new logic looks very robust!

Issue #51, opened 2024-03-07
`ark-cli`: External Crate for Printing Tables 


tareknaser:
We should consider using an external crate like [`radicle-term`](https://crates.io/crates/radicle-term) for printing tables, especially for `ark-cli list`. This can provide better customization options and simplify maintaining the codebase.

Issue #31, opened 2024-04-19
Enhancements for `fs-index` Crate


tareknaser:
**This is a tracking issue for brainstorming ideas to enhance `fs-index` crate**
---
- [x] `ResourceIndex` exposed API should be as discussed here https://github.com/ARK-Builders/arklib/issues/91#issuecomment-1998887042
- [x] `id_to_resources` should be of type `HashMap<HashType, Vec<IndexedResource>>` instead of `HashMap<ResourceId, CanonicalPathBuf>` because one id can map to multiple resource in case of collisions or same content 
- [x] Revise `ResourceIndex::update()` to rebuild the index and compare with the old index. Benchmarking for this feature should be conducted as per issue #30.
- [x] Implement file system watch and update index functionality as described in #21. 
> The difference between `watch_index` and `ark-cli monitor` is that `watch_index` uses `notify` to only update the index when a file system change is detected, as opposed to rebuilding the index from scratch on a time interval.
- [x] Create examples folder:
    - Include one example showing the normal scenario of creating a new index in a specified path and using the API.
    - Include another example showcasing index watch functionality.
- [x] Benchmark the performance of resource indexing queries and operations as outlined in issue #30.
- [x] Ensure that modules are private and only necessary structs and methods are exported.
- [x] Place unit tests in a separate file instead of `index.rs`.
- [x] Implement a custom `Serialize` implementation for `ResourceIndex` where double mappings are not written.

kirillt:
It seems that all sub-issues have been addressed.

Great work, thanks!

PR #67, opened 2024-05-24
Refactor `ark-cli` for Improved Code Organization and Readability

tareknaser:
## Description

Previously, `ark-cli` logic was concentrated in a few files, with large files such as `ark-cli/src/main.rs` containing all the logic for running the CLI by calling other functions. This made it difficult to debug, patch, and read the code. This PR is the first step towards addressing other issues within the crate.

## Changes
- Each command now resides in its own file, including its parameters and the function to run the logic.
- Each subcommand is placed in a folder, with each command as a separate file.
- All commands are included in a module that exports `pub enum Commands`.
- Defined a `run()` method to handle the logic for each command.
- The `main()` function now only calls `run()`, and if an error is encountered, it prints it to stderr. This enhances error handling.
- Avoided using `unwrap()` when applicable.
- Added `FIXME` notes to track planned changes.


ark-cli/src/commands/backup.rs
```diff
@@ -0,0 +1,91 @@
...
+#[derive(Clone, Debug, clap::Args)]
+#[clap(name = "backup", about = "Backup the ark managed folder")]
+pub struct Backup {
+    #[clap(value_parser, help = "Path to the root directory")]
+    roots_cfg: Option<PathBuf>,
+}
+
+impl Backup {
+    // FIXME: The logic for backup should be moved out of `ark-cli`
+    //        so that it can be used by other applications
```
kirillt, line 19: Good idea to make a crate from it, so developers could embed backup feature into their apps.

Let's create the issue for this.

ark-cli/src/commands/collisions.rs
```diff
@@ -0,0 +1,20 @@
...
+    #[clap(value_parser, help = "Path to the root directory")]
+    root_dir: Option<PathBuf>,
+}
+
+impl Collisions {
+    pub fn run(&self) -> Result<(), AppError> {
+        // FIXME: How does `monitor_index` handle `ark-cli collisions`?
+        monitor_index(&self.root_dir, None)
+    }
+}
```
kirillt, line 20: We might not need `collisions` command now because we have `list`.

We could implement these features for `list` command:
1. Reporting collisions number.
2. Reconstructing lists of all collided paths.
3. Grouping the collided paths by ids.

ark-cli/src/commands/link/utils.rs
```diff
@@ -38,6 +38,7 @@ pub fn load_link(
                 Err(AppError::LinkLoadError(format!(
                     "Path {:?} was requested. But id {} maps to path {:?}",
                     path,
+                    // FIXME: If `load_link` needs id, then it shouldn't be an optional parameter
                     id.clone().unwrap(),
                     path2,
                 )))
```
kirillt, line 44: An interesting question is how we can define app-specific resource kinds and commands related to these specific resource kinds, handling them uniformly?

That could be a plugin system. So, any app developer would create not only wrappers for each platform, but also CLI commands related to resource kinds this developer introduces. E.g. `.link` is custom file format defined by us and it contains `Link to Web` resource. It consists only of URL, but we also define some metadata and possible properties for it, e.g. domain and title.

ark-cli/src/commands/list.rs
```diff
@@ -0,0 +1,341 @@
...
+                    .map(|tags| tags.contains(filter))
+                    .unwrap_or(false)
+            });
+        }
+
+        let no_tags = "NO_TAGS";
+        let no_scores = "NO_SCORE";
+
+        // FIXME: We should use a table library to print this.
+        //        probably `comfy-table`
```
kirillt, line 183: We have such issue in this repo: https://github.com/ARK-Builders/ark-rust/issues/51

kirillt, line 183: We have such issue in this repo:
- #51

ark-cli/src/commands/monitor.rs
```diff
@@ -0,0 +1,22 @@
...
+    // FIXME: help message should specify what metric the interval is in
+    #[clap(help = "Interval to check for changes")]
+    interval: Option<u64>,
+}
+
+impl Monitor {
+    pub fn run(&self) -> Result<(), AppError> {
+        // FIXME: 1000 should be the default value in clap configuration
+        //        so users know
+        let millis = self.interval.unwrap_or(1000);
```
kirillt, line 19: Great idea, I think this could be done in this PR.

ark-cli/src/commands/render.rs
```diff
@@ -0,0 +1,42 @@
...
+            "low" => Ok(PDFQuality::Low),
+            _ => Err(AppError::InvalidRenderOption),
+        }?;
+        let buf = File::open(&filepath).map_err(|e| {
+            AppError::FileOperationError(format!("Failed to open file: {}", e))
+        })?;
+        let dest_path = filepath.with_file_name(
+            filepath
+                .file_stem()
+                // SAFETY: we know that the file stem is valid UTF-8 because it is a file name
```
kirillt, line 29: But why do we care about UTF-8 here? 🤔

ark-cli/src/commands/monitor.rs
```diff
@@ -0,0 +1,22 @@
...
+use crate::{monitor_index, AppError};
+
+#[derive(Clone, Debug, clap::Args)]
+#[clap(name = "monitor", about = "Monitor the ark managed folder")]
+pub struct Monitor {
+    #[clap(value_parser, help = "Path to the root directory")]
+    root_dir: Option<PathBuf>,
+    // FIXME: help message should specify what metric the interval is in
+    #[clap(help = "Interval to check for changes")]
+    interval: Option<u64>,
```
kirillt, line 12: Let's do it in this PR.

ark-cli/src/commands/backup.rs
tareknaser: Thanks. I have opened issue #68 to keep track of it.

ark-cli/src/commands/collisions.rs
tareknaser: In the planned changes for #31, I suggest handling collisions in a separate method within `ResourceIndex`. This approach will prevent us from managing collisions during each insert, update, or remove operation.

We can then modify the collisions command in `ark-cli` to display the output of the new method in a well-formatted table.


What do you think?

ark-cli/src/commands/link/utils.rs
tareknaser: Interesting idea. I created issue #69 to track this proposal. We might need more time to discuss and define a plan before starting implementation.

ark-cli/src/commands/list.rs
tareknaser: Right. Removed the FIXME

ark-cli/src/commands/render.rs
tareknaser: The "SAFETY" comment is there to justify the use of `unwrap()` since it's guaranteed to be safe in this case.

ark-cli/src/commands/monitor.rs
tareknaser: Sure. Updated all parameters with default values to be included in the Clap configuration.

ark-cli/src/commands/file/append.rs
```diff
@@ -35,12 +41,13 @@ impl Append {
             translate_storage(&Some(self.root_dir.to_owned()), &self.storage)
                 .ok_or(AppError::StorageNotFound(self.storage.to_owned()))?;
 
+        // FIXME: Why do we have `self.kind` and `self.storage`? Both are used to determine the storage type
```
kirillt, line 44: Kind is probably `file` or `folder`. It could be derived from names, at least for standard storages provided by ARK. But developers could also create their own storages which would be unknown to `ark-cli`, so the user would need to pass this information explicitly.

We could also figure out the storage type automatically by checking the path. But with atomic versions the path will always point to a folder, so we need to calculate depth of the folder. Folder storages would have depth of 3 (`storage/entry/version/value`), while file storage should have depth 2 (a file inside each version, i.e. `storage/version/table`).

ark-cli/src/commands/storage/list.rs
```diff
@@ -12,8 +12,9 @@ pub struct List {
     root_dir: Option<PathBuf>,
     #[clap(help = "Storage name")]
     storage: Option<String>,
-    #[clap(short, long, help = "TODO: what is this flag for?")]
-    versions: Option<bool>,
+    // TODO: What is the purpose of this field? Add a help message to clarify.
+    #[clap(short, long, action = clap::ArgAction::SetTrue)]
+    versions: bool,
```
kirillt, line 17: This must be about atomic versions.

ark-cli/src/commands/backup.rs
kirillt: Let's remove this fixme then.

ark-cli/src/commands/storage/list.rs
tareknaser: Thanks
Added a help message

ark-cli/src/commands/file/append.rs
tareknaser: > Kind is probably `file` or `folder`.

Right.
- `storage`: Required parameter of type `String`. Gets passed to `translate_storage()` to determine the file path and `storage_type`
- `kind`: Optional parameter of type `StorageType` and specifies if it's a file or folder. It overwrites the value for `storage_type`

Why would we need `kind` if `storage` already determines the storage type? Are there any cases where using `kind` would come in handy?

kirillt: The only real use-case for `kind` is custom storages, i.e. when the storage is not shipped by ARK framework, but the user implemented their own instance of `FileStorage` or `FolderStorage`. In this case, we can't figure it out by the path. But we still can figure it out by the folder structure as I've described above.

Overall, `kind` flag is more like a temporary helper. We should determine it automatically.

ark-cli/src/commands/storage/list.rs
```diff
@@ -0,0 +1,51 @@
...
+};
+
+#[derive(Clone, Debug, clap::Args)]
+#[clap(name = "list", about = "List resources in a storage")]
+pub struct List {
+    #[clap(value_parser, help = "Root directory of the ark managed folder")]
+    root_dir: Option<PathBuf>,
+    #[clap(help = "Storage name")]
+    storage: Option<String>,
+    #[clap(short, long, action = clap::ArgAction::SetTrue, help = "Whether or not to use atomatic versioning")]
```
kirillt, line 15: It's probably more like unfolding the versions in the table.. We should discuss this feature.

ark-cli/src/commands/storage/mod.rs
```diff
@@ -2,6 +2,8 @@ use clap::Subcommand;
 
 mod list;
 
+// FIXME: We should use new `fs-storage` crate to handle storage operations
```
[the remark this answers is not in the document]
tareknaser: Currently, `ark-cli` has its own storage module, defined [here](https://github.com/ARK-Builders/ark-rust/blob/main/ark-cli/src/models/storage.rs). This implementation handles two types of storage (files and folders) and includes methods such as `append`, `read`, and `insert`. It is primarily used to store information about tags, scores, stats, properties, metadata, previews, and thumbnails.

I think this functionality should be moved to a separate crate. We could either create a new crate or include it as a module in `fs-storage`.

This isn't a significant amount of work, but it's necessary, just like having a crate for backup and restore functionality is essential. `ark-cli` should focus on defining CLI structs only.

ark-cli/src/commands/file/append.rs
tareknaser: Got it.
I removed the FIXME since there is a reason to have both flags

ark-cli/src/commands/storage/mod.rs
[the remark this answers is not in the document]
tareknaser: I just noticed that #69 would likely provide all the functionality we need for `ark-cli storage` and in a generic way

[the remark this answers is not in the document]
kirillt: @tareknaser maybe, but #69 is a bit abstract, I think we need to remove current storage imitation from `ark-cli` and replace it with methods from `fs-storage`.

Can be done in this PR if it's small amount of work, otherwise create new issue for it please and remove the FIXME.

kirillt:
@tareknaser could you investigate where does this output come from? We need to register a bug, I guess.

```
temp % ark-cli list --link 
Loading app id at /Users/tareknasser/.ark...
http://google.com/                                                                                                                                                               
http://duckduckgo.com/                                                                                                                                                           
Username; Identifier;First name;Last name
booker12;9012;Rachel;Booker
grey07;2070;Laura;Grey
johnson81;4081;Craig;Johnson
jenkins46;9346;Mary;Jenkins
smith79;5079;Jamie;Smith
```

tareknaser:
> could you investigate where does this output come from? We need to register a bug, I guess.

I have traced this down and apparently, `ark-cli list --link` would attempt to read all resources and print the content of readable files without validation.

Here's the relevant code https://github.com/ARK-Builders/ark-rust/blob/main/ark-cli/src/main.rs#L161-L170 :
```rust
EntryOutput::Link => match File::open(path) {
		Ok(mut file) => {
			let mut contents = String::new();
			match file.read_to_string(&mut contents) {
				Ok(_) => (None, None, Some(contents)),
				Err(_) => return None,
			}
		}
		Err(_) => return None,
	},
```
To address this, I have pushed a commit to check if the file content is a valid link before printing it. This is a bit of a workaround.

ark-cli/src/commands/collisions.rs
kirillt: Let's remove the `FIXME` before merging. Feel free to create a follow-up issue if necessary.

ark-cli/src/commands/link/utils.rs
kirillt: Let's remove the `FIXME` before merging. We have generic #69 but specific link-related issue could be created, too.

kirillt:
@tareknaser

> To address this, I have pushed a commit to check if the file content is a valid link before printing it. This is a bit of a workaround.

If a follow-up issue is needed, feel free to register it. I think, we can merge the PR but please remove the last 2 `FIXME`s.

kirillt (approved):

PR #72, opened 2024-06-14
Remove release build from linux ci

twitu:
Remove linux release build as last step from github ci as it adds to build verification time.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -38,9 +38,6 @@ jobs:
       - name: Run tests
         run: cargo test --verbose
 
-      - name: Build Release
-        run: cargo build --verbose --release
-
   windows:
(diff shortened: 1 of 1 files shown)


tareknaser:
I don't see a strong reason to remove `cargo build --release` from our "verify build" CI workflow.

Currently, we don't have a dedicated release workflow, so this step could be a valuable sanity check. Down the road, we might even add `#[cfg(debug_assertions)]` which could disrupt our release code. 

The workflow itself completes in about 5 minutes, which seems reasonable. If anything, keeping it ensures we catch potential issues early on.

twitu:
Hmm I would argue against having a release build in a PR workflow it's better suited for scheduled perf tests. But it's not a pressing issue right now.

Issue #64, opened 2024-05-20
Add Conditional Collision Tracking Compilation Flag to `fs-index`


tareknaser:
## Description
Currently, `fs-index` keeps track of collisions regardless of the hash function used to compute the resource ID, whether it's cryptographic or not. 
This adds unnecessary computation in some cases. For cryptographic hash functions, there are no collisions, so this part can be ignored. 
> Note: some users might still want collision counting to track files with the same content.

## Plan
We plan to include a compilation flag in the `fs-index` crate to manage collision tracking. This was initially intended for #42, but we postponed it until we update the `ResourceIndex` API, which will simplify the collision tracking code. For more details, see [this comment](https://github.com/ARK-Builders/ark-rust/pull/42#issuecomment-2120833547).

## Notes
- The compilation flag might be better named `collision-counting` since it will now only track the number of occurrences of the same hash.
- An alternative approach could be to create a separate crate for the different implementation.

kirillt:
**Collision counting** (current implementation) might be useful but the developer need to know how to interpret them, and what to do when it's not just duplicates:
* with cryptographic hash function, collisions mean the data is duplicated
* with non-cryptographic hash function, collisions could mean both duplicates or real collisions

**Collision tracking** would be more useful for consumer apps, when used together with cryptographic hash functions, because we would also provide the developer with a list of duplicates for each resource id. For example, a photo app or documents vault could implement deduplication features using this.

**tl;dr:**
_collision counting_ is `ResourceId -> usize`
_collision tracking_ is `ResourceId -> Vec<PathBuf>`

Issue #20, opened 2024-03-18
`fs-index`: Store relative paths in `ResourceIndex` mappings


tareknaser:
## Description 
We should store entries paths relative to the root path of `ResourceIndex` 
### Current State
```rust
#[derive(PartialEq, Clone, Debug)]
pub struct ResourceIndex {
    pub id2path: HashMap<ResourceId, CanonicalPathBuf>,
    pub path2id: HashMap<CanonicalPathBuf, IndexEntry>,

    pub collisions: HashMap<ResourceId, usize>,
    root: PathBuf,
}
```
- We are currently storing the canonical path of each entry in the mappings. 
- `ResourceIndex.root` is already the canonical root path 
### Proposal
- Modify `ResourceIndex` to store entries' paths relative to the root (in `id2path` and `path2id`)
- Expose a method to reconstruct the canonical path of an entry in the index by prepending it with the root path

tareknaser:
This was addressed in #79

Issue #16, opened 2024-03-15
`fs-index`: Benchmarking Resource Index API 


tareknaser:
## Description
Benchmark the Resource Index API while experimenting with alternative data structures for the `ResourceIndex` implementation. 

### Current State
Currently, the `ResourceIndex` struct has a design with double hashmaps:
```rust
#[derive(PartialEq, Clone, Debug)]
pub struct ResourceIndex {
    pub id2path: HashMap<ResourceId, CanonicalPathBuf>,
    pub path2id: HashMap<CanonicalPathBuf, IndexEntry>,

    pub collisions: HashMap<ResourceId, usize>,
    root: PathBuf,
}
```
The current `ResourceIndex` isn't serialized directly and loaded due to the use of double hashmaps. Instead, we manually store the resources and parse the information again when loading.
### Proposals
One of the proposals is this:
```rust
#[derive(PartialEq, Clone, Debug, Serialize, Deserialize)]
pub struct IndexedResource {
    pub id: ResourceId,
    pub path: PathBuf,
    pub last_modified: SystemTime,
}

#[derive(PartialEq, Clone, Debug, Serialize, Deserialize)]
pub struct ResourceIndex {
    pub resources: Vec<IndexedResource>,
    pub root: PathBuf,
}
```
### Motivation
To assess the performance impact of other data structures, we should add benchmarks that involve storing entries in the resource index and then querying them using `get_resource_by_id` or `get_resource_by_path` methods.

This way, we can evaluate the efficiency of different data structures regarding insertion, retrieval, and overall performance of the Resource Index API.

tareknaser:
I think this was addressed in #79

PR #79, opened 2024-06-30
Refactor `ResourceIndex` and Associated Methods

tareknaser:
## Description
This PR outlines a series of improvements and refactors to the `ResourceIndex` and its associated methods, as discussed in [this comment](https://github.com/ARK-Builders/arklib/issues/91#issuecomment-1998887042).
## Changes
1. Refactor `ResourceIndex` API to align with the proposed changes discussed in the linked comment.
2. Move unit tests to a separate file from `index.rs`.
3. Implement a custom Serialize implementation for ResourceIndex that avoids writing double mappings.
4. Revise `ResourceIndex::update_all()` to rebuild the index and compare it with the old index.
5. Update `id_to_resources` to be of type `HashMap<HashType, Vec<IndexedResource>>` instead of `HashMap<ResourceId, CanonicalPathBuf>` to handle cases where one ID maps to multiple resources due to collisions or identical content.

**Related Issues:**
- #64: Collision Counting
- #20, #16, #31: Various enhancements and fixes related to the index functionality
### Additional Changes:
- Define new rules for `rustfmt` to wrap long comments.
- Use the nightly build in CI for `rustfmt` experimental rules.
- Change `target_os` to `target_family`.
## Todo:
- [x]  Update `ark-cli` crate to reflect the changes.
- [x]  Add examples for `fs-index`.
- [x]  Write a simple `README` for `fs-index`.
- [x] Add a method for collision counting as discussed in #64.
- [x] Make sure not to index empty files.


fs-index/src/tests.rs
```diff
@@ -0,0 +1,699 @@
+use std::{fs, path::Path};
+
+use anyhow::{anyhow, Result};
+use tempfile::TempDir;
+
+use data_resource::ResourceId;
+use dev_hash::Crc32;
```
kirillt, line 7: We should have tests using cryptographic hash function as well.

I see two approaches:
1. Copy-paste and adjust, so we'd have 2 files: `tests_crc32.rs` and `tests_blake3.rs`.
2. Write simple automation function/macro and convert existing test cases into automated test cases, where we iterate through vector of hash functions (2 items is enough but could be more) and run the test case itself inside the loop. We could also define 2 vectors: cryptographic and non-cryptographic, some test cases should be written differently, on top of these vectors.

Approach 1 will reduce maintainability of the codebase, while approach 2 is probably overkill and could also decrease readability. Probably, we could have something in between?

fs-index/src/index.rs
```diff
@@ -1,1120 +1,501 @@
...
-        .unwrap_or(false)
+/// Represents the result of an update operation on the ResourceIndex
+#[derive(PartialEq, Debug)]
+pub struct IndexUpdate<Id: ResourceId> {
+    /// Resources that were added during the update
+    added: Vec<IndexedResource<Id>>,
+    /// Resources that were modified during the update
+    modified: Vec<IndexedResource<Id>>,
+    /// Resources that were removed during the update
+    removed: Vec<IndexedResource<Id>>,
```
kirillt, line 146: Is `last_modified` timestamp correct in items of `removed` vector?

It's time of last modification before deletion, not time of removal, right?

fs-index/src/index.rs
```diff
@@ -1,1120 +1,501 @@
...
+        let mut removed = Vec::new();
+
+        let new_index = ResourceIndex::build(&self.root)?;
+
+        // Compare the new index with the old index
+        let current_resources = self.resources();
+        let new_resources = new_index.resources();
+        for resource in new_resources.clone() {
+            // If the resource is in the old index, check if it has been
+            // modified
```
kirillt, line 326:
```suggestion
            // If the resource is in the old index,
            // check if it has been modified
```

fs-index/src/index.rs
```diff
@@ -1,1120 +1,501 @@
...
+        let path = relative_path.as_ref();
+        let full_path = self.root.join(path);
+        if !full_path.exists() {
+            return Err(ArklibError::Path(format!(
+                "File does not exist: {:?}",
+                full_path
+            )));
+        }
+        let metadata = fs::metadata(&full_path)?;
+        // return an error if the file is empty
```
kirillt, line 384:
```suggestion
        // empty files don't have content, so we can't compute id
```

fs-index/src/index.rs
```diff
@@ -1,1120 +1,501 @@
...
-        .to_str()
-        .map(|s| s.starts_with('.'))
-        .unwrap_or(false)
+/// Represents the result of an update operation on the ResourceIndex
+#[derive(PartialEq, Debug)]
+pub struct IndexUpdate<Id: ResourceId> {
+    /// Resources that were added during the update
+    added: Vec<IndexedResource<Id>>,
+    /// Resources that were modified during the update
+    modified: Vec<IndexedResource<Id>>,
```
kirillt, line 144: Am I right that modification is an update of content by some path?

If a resource changes its location, is it represented as removal and addition at the same time?

fs-index/src/index.rs
```diff
@@ -1,1120 +1,501 @@
...
-                ResourceIndex::build(path.clone());
+    /// Update the index with the latest information from the file system
+    pub fn update_all(&mut self) -> Result<IndexUpdate<Id>> {
+        log::debug!("Updating index at root path: {:?}", self.root);
+
+        let mut added = Vec::new();
+        let mut modified = Vec::new();
+        let mut removed = Vec::new();
+
+        let new_index = ResourceIndex::build(&self.root)?;
```
kirillt, line 319: We should not throw away the previous definition. This implementation is a lot simpler, which is good. But the older implementation was optimized. We should benchmark both functions.

fs-index/src/index.rs
```diff
@@ -1,1120 +1,501 @@
...
+                self.get_resource_by_path(&resource.path)
+            {
+                if current_resource != &resource {
+                    modified.push(resource.clone());
+                }
+            }
+            // If the resource is not in the old index, it has been added
+            else {
+                added.push(resource.clone());
+            }
```
kirillt, line 337: The `resource.path` is not found in the index, but a resource with same content can be already tracked by another path. The app developer should treat additions only if the resource didn't exist before.

Probably we should implement "moved" event instead of "modified". Resources cannot be modified because they are identified by their content, if the content is changed then it's another resource. But resources can move, i.e. change their paths.

fs-index/src/index.rs
```diff
@@ -1,1120 +1,501 @@
...
+    /// # Arguments
+    /// * `relative_path` - The path of the file to be added (relative to the
+    ///   root path of the index).
+    ///
+    /// # Returns
+    /// Returns `Ok(resource)` if the file was successfully added to the index.
+    ///
+    /// # Errors
+    /// - If the file does not exist in the file system.
+    /// - If there was an error calculating the checksum of the file.
```
kirillt, line 368: The older comment:
```
 354     /// The caller must ensure that:
 355     /// * the index is up-to-date except this single path
 356     /// * the path hasn't been indexed before
 357     ///
 358     /// Should only be used if reactive updates are not possible.
```

However, now I think we should care about id, not path. The use-case is an app that created resource, and it's very unlikely to have this resource indexed already (user created new image for instance). Then we can just add the resource to the index, skipping the whole index update.

So, the caller must ensure that:
1. The index is up-to-date except this single id.
2. The id hasn't been indexed before.

All `track` functions are intended to be used by app developers for optimized index manipulations, but they should understand the circumstances.

fs-index/src/index.rs
```diff
@@ -1,1120 +1,501 @@
...
+    /// * `relative_path` - The relative path of the file to be modified.
+    ///
+    /// # Returns
+    /// Returns `Ok(new_resource)` if the resource was successfully modified in
+    /// the index.
+    ///
+    /// # Errors
+    /// - If there was a problem removing the old resource from the index.
+    /// - If there was a problem adding the new resource to the index.
+    pub fn track_modification<P: AsRef<Path>>(
```
kirillt, line 472: Original motivation was here:
- https://github.com/ARK-Builders/arklib/issues/38

The use-case was updating the resource in one app, after its modification in another app.

But now I think, this function is not needed, because:
1. The editor app can have different flow, like opening a resource by one path, and saving the modified version to another path. Reactive API is much more robust here, especially with `notify`. We should ensure that `update_all` is efficient, too. We had a bug in there causing very slow re-indexing every time a single update was made. That's why the `track_update` appeared originally.
2. If the editor overwrites the resource by same path, then it still can use combination of `track_deletion` and `track_addition`.

dev-hash/benches/crc32.rs
```diff
@@ -17,7 +17,8 @@ fn generate_random_data(size: usize) -> Vec<u8> {
     (0..size).map(|_| rng.gen()).collect()
 }
 
-/// Benchmarks the performance of resource ID creation from file paths and random data.
+/// Benchmarks the performance of resource ID creation from file paths and
+/// random data.
```
kirillt, line 21:
```suggestion
/// Benchmarks the performance of resource ID creation
///  from file paths and random data.
```

dev-hash/benches/blake3.rs
```diff
@@ -17,7 +17,8 @@ fn generate_random_data(size: usize) -> Vec<u8> {
     (0..size).map(|_| rng.gen()).collect()
 }
 
-/// Benchmarks the performance of resource ID creation from file paths and random data.
+/// Benchmarks the performance of resource ID creation from file paths and
+/// random data.
```
kirillt, line 21:
```suggestion
/// Benchmarks the performance of resource ID creation
///  from file paths and random data.
```

data-resource/src/lib.rs
```diff
@@ -3,16 +3,16 @@
...
 ///
-/// Resources are identified by a hash value, which is computed from the resource's data.
-/// The hash value is used to uniquely identify the resource.
+/// Resources are identified by a hash value, which is computed from the
+/// resource's data. The hash value is used to uniquely identify the resource.
 ///
-/// Implementors of this trait must provide a way to compute the hash value from the resource's data.
+/// Implementors of this trait must provide a way to compute the hash value from
+/// the resource's data.
 pub trait ResourceId:
```
kirillt, line 16:
```suggestion
/// Implementors of this trait must provide a way to compute
///  the hash value from the resource's data.
pub trait ResourceId:
```

ark-cli/src/index_registrar.rs
```diff
@@ -34,7 +36,9 @@ pub fn provide_index<P: AsRef<Path>>(
     }
 
     log::info!("Index has not been registered before");
-    match ResourceIndex::provide(&root_path) {
+    // If the index has not been registered before, we need to load it, update
+    // it and register it
```
kirillt, line 40:
```suggestion
    // If the index has not been registered before,
    // we need to load it, update it and register it
```

ark-cli/src/commands/render.rs
```diff
@@ -26,7 +26,8 @@ impl Render {
         let dest_path = filepath.with_file_name(
             filepath
                 .file_stem()
-                // SAFETY: we know that the file stem is valid UTF-8 because it is a file name
+                // SAFETY: we know that the file stem is valid UTF-8 because it
+                // is a file name
```
kirillt, line 30:
```suggestion
                // SAFETY: we know that the file stem is valid UTF-8
                // because it's a file name
```

ark-cli/src/commands/list.rs
```diff
@@ -123,21 +124,18 @@ impl List {
...
+                    EntryOutput::Path => (Some(path.to_owned()), None, None),
+                    EntryOutput::Id => (None, Some(id.to_owned()), None),
                     EntryOutput::Link => match File::open(path) {
                         Ok(mut file) => {
                             let mut contents = String::new();
                             match file.read_to_string(&mut contents) {
                                 Ok(_) => {
-                                    // Check if the content of the file is a valid url
+                                    // Check if the content of the file is a
+                                    // valid url
```
kirillt, line 138:
```suggestion
                                    // Check if the content
                                    // of the file is a valid url
```

kirillt (changes requested):
Mostly good, but `update_all` might be different from what it was. Not sure about `modified` field in the `IndexUpdate` struct. Overall, the index should provide API in terms of resource, not paths. Ideally, the user should not care about locations, only about content.

Also, let's move Track API into separate PR because it should be considered more carefully.

fs-index/src/index.rs
tareknaser: It's the time of the last modification.

The `IndexedResource` fields are only updated when we call the `update_all()` or `track_*()` methods.

fs-index/src/index.rs
tareknaser: The logic for tracking modifications is to first track removal and then track addition. This is to make sure that the ID and last modified time are updated.

> If a resource changes its location, is it represented as removal and addition at the same time?

Yes, if a resource changes its path, it is treated as an addition and removal. This is because we use paths as the unique identifier for file systems. Hashes, including cryptographic hashes, can be the same if the content is identical.

fs-index/src/index.rs
tareknaser: > The `resource.path` is not found in the index, but a resource with same content can be already tracked by another path.

This suggests that we might want to use the file hash as the unique identifier. So far, we've been using paths to determine if it's the same file due to the file system making sure of no duplicate paths.

To use the resource ID as the sole unique identifier, we would need to prevent adding two files with the same content. Is this acceptable?

fs-index/src/tests.rs
tareknaser: I added a conditional import statement, so we don't need to duplicate the entire test suite. This way, we also avoid the need for additional automation functions or macros.

I've included running the tests with both hash functions in our CI workflows.

Let me know what you think of b3eeda0f3e52a1055c031e667cf847dc68925385

fs-index/src/index.rs
tareknaser: I wrote benchmarks to be sure that the proposed logic changes don't impact performance. The `resource_index_benchmark` includes an `index_update_all` function, which:
- Creates 50 files
- Builds the index
- Creates 50 new files, removes 30 files, and modifies 10 files
- Updates the index
I didn't notice any performance drop from the changes proposed in the PR.

This result isn't surprising since the old implementation doesn't seem so different in the core. In the main branch:
- `ResourceIndex::build()` calls `discover_paths()` and `scan_entries()` [here](https://github.com/ARK-Builders/ark-core/blob/main/fs-index/src/index.rs#L53-L54), each once.
- `ResourceIndex::update_all()` calls the same functions plus extra logic:
    - It calls `discover_path()` [here](https://github.com/ARK-Builders/ark-core/blob/main/fs-index/src/index.rs#L207)
    - It calls `scan_entries()` [here](https://github.com/ARK-Builders/ark-core/blob/main/fs-index/src/index.rs#L316) and [here](https://github.com/ARK-Builders/ark-core/blob/main/fs-index/src/index.rs#L320)

Additionally, the code quality improvement is significant imho. The current code for `ResourceIndex::update_all()` is large, error-prone, and difficult to debug. The proposed changes are simpler and easier to debug, with minimal chances of error.

----

To be sure, I pushed similar code to a branch in my fork: [GitHub link](https://github.com/ARK-Builders/ark-core/commit/266347db29e4a899a5cd6a927af9ff1bc0cec527). Benchmarking this against the proposed code gives the following results:
```shell
resource_index/index_update_all
                        time:   [11.218 ms 11.258 ms 11.303 ms]
                        change: [+0.2740% +0.7508% +1.2455%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 6 outliers among 100 measurements (6.00%)
  4 (4.00%) high mild
  2 (2.00%) high severe
```
This shows no noticeable performance hit.


What do you think?

fs-index/src/tests.rs
kirillt: Not bad, but I think with this approach we might run into problems later.

I think we should use approach 2 described above. The macro can be defined in utils, that takes a specific function and runs some computation with this function.

Then our tests could be split into 3 files:
* `tests_all.rs` for testing invariants that must hold regardless of the type of hash function
* `tests_crypto.rs` for testing invariants that hold true only for cryptographic hash functions
* `tests_fast.rs` for testing invariants that hold true only for non-cryptographic functions

So, each unit test would look like:
```
with_hash_function([blake3, crc32, sha256]) {
...
}
```

or separate macros for hash function types:
```
with_crypto_hashing {
...
}

with_fast_hashing {
...
}
```

This way it would be easier to control what functions are used in each test.

fs-index/src/index.rs
kirillt: I don't think that we should provide users with this timestamp for removed items. It might be confusing. But `modified` entry is confusing to me in either case, because resources are immutable. They can be only added or removed, their locations are secondary. We might support `moved` event though.

In future versions of the API, I would expect events like `new version` for cases when one resource is derived from another.

fs-index/src/index.rs
kirillt: I'm not sure that the two definitions are semantically equivalent.

fs-index/src/index.rs
kirillt: The high-level goals are:
1. to find all different resources
2. determine their paths
3. detect if we have different paths for a single resource
4. list all different paths for same a single resource

1 and 2 are mandatory goals, 3 and 4 are optional goals.

kirillt:
By the way, let's use squashing and rebasing only before the merge into `main`.

Otherwise, it's difficult to see what was changed after last review...

fs-index/src/tests.rs
tareknaser: I agree, this is a better approach. 

My initial suggestion might have been confusing `fs-index` users, as the feature flags don’t add value to the library itself. Instead, I defined a macro `generate_tests_for_hash` to generate tests for a given hash type (implementing `ResourceId`). This macro is used to run tests for both `Blake3` and `CRC32`, and it can be extended for any hash type from the `dev_hash` crate.

The module organization is clear. Here’s how the tests look in the terminal:
```shell
running 22 tests
test tests::blake3::test_build_index_with_file ... ok
test tests::blake3::test_build_index_with_directory ... ok
test tests::blake3::test_build_index_with_empty_file ... ok
test tests::blake3::test_build_index_with_multiple_files ... ok
test tests::blake3::test_hidden_files ... ok
test tests::blake3::test_build_index_with_multiple_directories ... ok
test tests::crc32::test_build_index_with_directory ... ok
test tests::crc32::test_build_index_with_file ... ok
test tests::crc32::test_build_index_with_empty_file ... ok
test tests::blake3::test_add_colliding_files ... ok
test tests::blake3::test_store_and_load_index ... ok
test tests::crc32::test_add_colliding_files ... ok
test tests::blake3::test_resource_index_update ... ok
test tests::blake3::test_num_collisions ... ok
test tests::crc32::test_hidden_files ... ok
test tests::crc32::test_build_index_with_multiple_files ... ok
test tests::blake3::test_store_and_load_index_with_collisions ... ok
test tests::crc32::test_build_index_with_multiple_directories ... ok
test tests::crc32::test_store_and_load_index ... ok
test tests::crc32::test_num_collisions ... ok
test tests::crc32::test_resource_index_update ... ok
test tests::crc32::test_store_and_load_index_with_collisions ... ok

test result: ok. 22 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.02s
```
I chose to keep all test modules in one file for simplicity. Each module for a specific hash function is only a few lines, so it keeps the file concise.

Is this fine or do you think each of the modules should be in a separate file?

fs-index/src/index.rs
tareknaser: Changed the return value of `ResourceIndex::update_all()` to include only two maps: added and removed.

fs-index/src/index.rs
tareknaser: Updated the logic for the `update_all()` method. Let me know what you think.

fs-index/src/index.rs
```diff
@@ -1,1120 +1,436 @@
...
+        };
+        Ok(index)
     }
 
-    #[test]
-    fn update_one_should_index_delete_file_successfully() {
-        run_test_and_clean_up(|path| {
-            create_file_at(path.clone(), Some(FILE_SIZE_1), Some(FILE_NAME_1));
+    /// Update the index with the latest information from the file system
+    pub fn update_all(&mut self) -> Result<IndexUpdate<Id>> {
```
kirillt, line 284: @kirillt compare with the original implementation

fs-index/src/tests.rs
```diff
@@ -0,0 +1,539 @@
...
+    generate_tests_for_hash!(Blake3);
+}
+
+mod crc32 {
+    use dev_hash::Crc32;
+
+    use crate::generate_tests_for_hash;
+
+    generate_tests_for_hash!(Crc32);
+}
```
kirillt, line 30: I think we can improve these tests even further. Core ideas:
- We should not declare separate modules for different hash functions
- We should have more finely-grained control in test cases

----

However, I'm not sure if we can declare lists of hash functions statically in the way that it would be possible to iterate them in test cases. We probably still can use a macro.

- [ ] Define a function, or a macro parameterized by many functions at once, which runs the code section inside, for every function passed to it: `for_each(crc32, blake3, sha256, whatever) { /* test case */ }`

    - If it's a simple for loop through a vector like `for f in vec![crc32, blake3, whatever] { /* test case */ }` that'd be even better, because the code inside the loop could be developed without commenting-out the loop itself. But I'm afraid that it's too difficult to do in Rust
    
- [ ] Define two broad categories of "cryptographic" and "fast" hashes as plain vectors, or as a 2nd-level macros (wrappers), so we can simply add new functions into one of the two categories, and the test cases will work with new functions automatically without necessity to update them manually.

    - This will allow us to describe invariants in generic terms, not specific functions
    ```
    fn test_no_collisions_with_cryptographic_hashing() {
        for_cryptographic_hashing! {
            // setup a folder without duplicates
            // index it
            assert_eq!(index.collisions, 0);
        }
    }
    ```

    - But we are still free to create custom test cases for one or two specific functions, these test cases will not be run excessively for functions that are not involved
    ```
    fn test_no_collisions_with_cryptographic_hashing() {
        for_each_hashing!(take_first_byte, take_last_byte) {
            // setup a folder without duplicates
            // index it
            assert_eq!(index.keys.size(), 8); //synthetic example
        }
    }
    ```

tareknaser: Thanks for the detailed review

I've defined a macro `hash_tests` to generate tests for function and hash type pairs, and used it to run tests for the two hash functions currently supported by `dev-hash`.

Adding a new test is straightforward: just add an entry in the macro call with the test function and hash function you want to use.

I've also included instructions on how to navigate the test code and add new tests.

We could use `rstest` to [parameterize tests](https://github.com/la10736/rstest?tab=readme-ov-file#parametrize), but it might not improve code quality much and could add complexity. The current code is well-documented and easy to understand.


Let me know if that's what you had in mind

fs-index/src/index.rs
tareknaser: I looked into the CI error for Windows in this PR and found that the test case also fails for Windows in `arklib`. It seems the logic in `update_all()` needs another review and debugging to understand why it only fails on Windows.

You can see the CI results for `arklib` from my fork here:
- [Comparison with main](https://github.com/ARK-Builders/arklib/compare/main...tareknaser:arklib:main)
- [CI run details](https://github.com/tareknaser/arklib/actions/runs/10011315002)

fs-index/src/index.rs
```diff
@@ -1,1120 +1,436 @@
...
-                .expect("Should remove file successfully");
+        // `updated_entries` is the difference between preserved_entries and
+        // current_entries where the last modified time has changed
+        // significantly
+        let updated_entries: HashMap<PathBuf, IndexedResource<Id>> =
+            preserved_entries
+                .iter()
+                .filter_map(|(path, resource)| {
+                    if current_entries.contains_key(path) {
+                        None
```
kirillt, line 332: Is this correct?

fs-index/src/index.rs
```diff
@@ -1,1120 +1,436 @@
...
+                    } else {
+                        let our_entry =
+                            self.path_to_resource.get(path).unwrap();
```
kirillt, line 335: Here and in many places in this file, expressions like `map.get(key).unwrap()` can be simplified to just `map[key]`.

fs-index/src/index.rs
kirillt: Let's also use `HashSet` instead of `Vec` to ensure that duplicates are not included. Order of elements should not matter.

kirillt: And we don't need to provide users with `last_modified` timestamps of removed resources, because they are outdated and do not reflect time of removal. So it's confusing and may provoke misuse.

Also, we should not provide users with paths of removed resources. There is no use for them. If we keep paths of removed resources in our API, users could think that they should update elements by these paths. But this is job for `fs-index`, users only need to track resource ids, read content by path when it is updated (but not when removed) and use timestamps for sortings.

So, we should only emit ids of removed resources.
```
pub struct IndexUpdate<Id: ResourceId> {
    /// Resources that were added during the update
    added: HashSet<IndexedResource<Id>>,
    
    /// Resources that were removed during the update
    removed: HashSet<Id>,
}
```

kirillt: Lastly, we should maintain invariant that `added` set never contains elements with same ids and different paths/timestamps. This can be expressed in the code:
```
pub struct IndexEntry {
    /// The path of the resource, relative to the root path
    path: PathBuf,

    /// The last modified time of the resource (from the file system metadata)
    last_modified: SystemTime,
}

pub struct IndexUpdate<Id: ResourceId> {
    /// Resources that were added during the update
    added: HashMap<Id, IndexEntry>,

    /// Resources that were removed during the update
    removed: HashSet<Id>
}
```

fs-index/src/index.rs
```diff
@@ -1,1120 +1,436 @@
...
-use data_error::{ArklibError, Result};
+use data_error::Result;
 use data_resource::ResourceId;
 use fs_storage::{ARK_FOLDER, INDEX_PATH};
 
-#[derive(Eq, Ord, PartialEq, PartialOrd, Hash, Clone, Debug)]
-pub struct IndexEntry<Id: ResourceId> {
-    pub modified: SystemTime,
-    pub id: Id,
-}
```
kirillt, line 22: This structure should be brought back, see the comment about invariant and `added: HashMap`

fs-index/src/index.rs
```diff
@@ -1,1120 +1,436 @@
...
-            }
-        })
-        .collect()
+    /// The root path of the index (canonicalized)
+    pub(crate) root: PathBuf,
+    /// A map from resource IDs to resources
+    ///
+    /// Multiple resources can have the same ID (e.g., due to hash collisions
+    /// or files with the same content)
+    pub(crate) id_to_resources: HashMap<Id, Vec<IndexedResource<Id>>>,
```
kirillt, line 118: We should remove ids from the right side of the mapping:
```
id_to_paths: HashMap<Id, HashSet<IndexEntry>>
```

If it's possible to also remove timestamps from the right side of the mapping, without breaking `update_all`, then it would be even better:
```
id_to_paths: HashMap<Id, HashSet<Path>>
```

fs-index/src/index.rs
kirillt: Mostly looks good. I suspect there is one bug though (missed negation).

By the way, in benchmarks we should have thousands of files, not dozens.

fs-index/src/tests.rs
kirillt: I've seen `rstest` before, good thing but we don't need it right now.

The way how you've implemented works, but curious if you've tried my suggestions above. Does it result in less maintainable codebase?

fs-index/src/index.rs
kirillt, line 332: `None` looks incorrect here.. We need a test case for this. In the original implementation, the condition was negated: `!curr.contains` what makes sense because updated paths are preserved and current.

And comment tells "difference", but should it be "intersection"?

fs-index/src/tests.rs
tareknaser: I attempted the first proposed approach. The following dummy code to illustrate:
```rust
/// A macro to take many functions and run them as tests for crc32 and blake3
#[macro_export]
macro_rules! hash_tests_crc32_and_blake3 {
    ($($name:ident,)*) => {
        $(
            for hash_type in &[Crc32, Blake3] {
                #[test]
                fn $name() {
                    $name::<$hash_type>();
                }
            }
        )*
    };
}

hash_tests_crc32_and_blake3!(
    test_store_and_load_index,
    test_store_and_load_index_with_collisions,
    test_build_index_with_file,
    test_build_index_with_empty_file,
    test_build_index_with_directory,
    test_build_index_with_multiple_files,
    test_build_index_with_multiple_directories,
    test_resource_index_update,
    test_add_colliding_files,
    test_num_collisions,
    test_hidden_files,
);
```

This fails to compile with error: 
```shell
macro expansion ignores token `for` and any following
the usage of `hash_tests_crc32_and_blake3!` is likely invalid in item context
```

Attempting to replace it with:
```rust
macro_rules! hash_tests_crc32_and_blake3 {
    ($($name:ident,)*) => {
        $(
            #[test]
            fn $name() {
                $name::<Crc32>();
            }

            #[test]
            fn $name() {
                $name::<Blake3>();
            }
        )*
    };
}
```
or 
```rust
macro_rules! hash_tests_crc32_and_blake3 {
    ($($name:ident,)*) => {
        $(
            #[test]
            fn test_for_crc32() {
                $name::<Crc32>();
            }

            #[test]
            fn test_for_blake3() {
                $name::<Blake3>();
            }
        )*
    };
}
```
fails to compile because multiple test functions have the same name, which Rust doesn't allow. The only way to avoid this error is by separating them into different modules, as in the previous code, but we want to avoid that.

Additionally, we can't dynamically set function names based on the hash type because function names in Rust don't exist at runtime.

While debugging this, I realized we could simply specify the name of the function to the macro, so it's clear. This is the current approach in this PR. I think this gives us fine-grained control over which tests to run and with which hash functions.

fs-index/src/index.rs
tareknaser: Updated this one. Other places in the code use `get_mut()` to get a mutable reference from the map.

fs-index/src/index.rs
tareknaser: > And comment tells "difference", but should it be "intersection"?

Right. thanks for catching that.
Updated in d4d7e73

fs-index/src/index.rs
tareknaser: > By the way, in benchmarks we should have thousands of files, not dozens.

For all `ReourceIndex` methods benchmarks or just for `update_all()` ?

tareknaser: > By the way, in benchmarks we should have thousands of files, not dozens. 

I agree. just updated the benchmarks temp directory to contain 10,000 files. For `ResourceIndex::update_all()`, we create 5,000 new files, remove 3,000 files, and modify 1,000 files.

fs-index/src/index.rs
tareknaser: > users only need to track resource ids, read content by path when it is updated

So, I assume we need `IndexUpdate` to be:
```rust
/// Represents the result of an update operation on the ResourceIndex
#[derive(PartialEq, Debug)]
pub struct IndexUpdate<Id: ResourceId> {
    /// Resources that were added during the update
    added: HashMap<Id, IndexedResource<Id>>,
    /// Resources that were removed during the update
    removed: HashSet<Id>,
}
```
Where `added` is a map between `Id` and `IndexedResource` (Id, path, last modified time), not `IndexEntry` (Id, last modified time).
right?

fs-index/src/index.rs
tareknaser: I'm not sure where `IndexEntry` should be used. If it's for the `added` map in `IndexUpdate`, the user wouldn't have access to the paths of the added resources. I think we'd want this information to be available in our API.

fs-index/src/index.rs
tareknaser: Updated `ResourceIndex` to include the following mappings:
```rust
#[derive(Clone, Debug)]
pub struct ResourceIndex<Id>
where
    Id: Eq + Hash,
{
    /// The root path of the index (canonicalized)
    pub(crate) root: PathBuf,
    /// A map from resource IDs to paths
    ///
    /// Multiple resources can have the same ID (e.g., due to hash collisions
    /// or files with the same content)
    pub(crate) id_to_paths: HashMap<Id, HashSet<PathBuf>>,
    /// A map from resource paths to resources
    pub(crate) path_to_resource: HashMap<PathBuf, IndexedResource<Id>>,
}
```

Updated the custom serialization and deserialization functions accordingly. Also made changes to `update_all()` to reflect these updates.

tareknaser: I have a question regarding `update_all()`. We currently have two mappings:
- path -> resource (id, path, modified time)
- id -> path

We allow multiple resources to have the same ID (either due to the same content or hash collisions in non-cryptographic hashes).

Here is the return type of `update_all()`:
```rust
/// Represents the result of an update operation on the ResourceIndex
#[derive(PartialEq, Debug)]
pub struct IndexUpdate<Id: ResourceId> {
    /// Resources that were added during the update
    added: HashMap<Id, IndexedResource<Id>>,
    /// Resources that were removed during the update
    removed: HashSet<Id>,
}
```
If we have an index where three resources map to the same ID and one of these resources is removed, should we return the ID as if it was removed? It might be confusing since the ID still exists in the index.

kirillt: @tareknaser good question. Please, register new issue for this. In the opposite scenario, when new duplicate appeared, I would expect that we mention it in `added` field, but not in `removed` as we do with renamings. But if a duplicate is removed, we might need special field in the `IndexUpdate` structure.

kirillt: By the way, I think `added` field should be a mapping into collections of paths:
```
added: HashMap<Id, HashSet<PathBuf>>
```

On other hand, providing users with the timestamps could be also a good idea..

fs-index/src/index.rs
kirillt: No, it was used in the `ResourceIndex` structure, in the `path_to_resource` part.

fs-index/src/tests.rs
kirillt: @tareknaser could you try this approach? https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=7399c2540956391e1265469e2c2d693e

Ideally, we should be able just iterate through types implementing the trait `ResourceId`. Without generating the actual tests from templated functions.

tareknaser: Yes, this compiles fine and does what we need. Let me know what you think of commit d076f34

fs-index/src/index.rs
tareknaser: > On other hand, providing users with the timestamps could be also a good idea..


I think we should include the modified time of added resources since this covers both newly created and modified resources.

For modified resources, users can compare the path of the updated resource with the resource of the same path from the previous state. The modified time can help users in some cases make informed decisions.

kirillt: @tareknaser this sounds like what our library should do, not users. But timestamps could be useful for sorting resources by date, so let's keep them for the moment.

fs-index/src/tests.rs
kirillt: Looks good now, thanks.

kirillt:
Actually, we need multiple values in the `added` field... because we can have collisions/duplicates, so better to inform the user about it:
```
pub struct ModifiedPath {
    path: PathBuf,
    modified: SystemTime
}

pub struct IndexUpdate<Id: ResourceId> {
    added: HashMap<Id, HashSet<UpdatedPath>>,
    removed: HashSet<Id>,
}
```

And we should not duplicate `ResourceId` in left and right sides, otherwise we'd need to maintain equality of them. So, auxiliary structure `ModifiedPath` is handy.

Last actions points before merging:
- [ ] multiple values in `added` field of `IndexUpdate`
- [ ] fix the build

tareknaser:
> Actually, we need multiple values in the `added` field... because we can have collisions/duplicates, so better to inform the user about it:

Thanks for catching that.
I've made the updates and added a unit test to verify the implementation's correctness. The test involves updating the content of a file and adding a new file with the same content, then confirming that `update_all()` returns two entries for the new ID. 
I also included some additional tests to cover corner cases we hadn't previously checked with `update_all()`.


> And we should not duplicate `ResourceId` in left and right sides, otherwise we'd need to maintain equality of them. So, auxiliary structure `ModifiedPath` is handy.

I renamed `IndexEntry` to `ResourceIdWithTimestamp` and created a separate struct called `ResourcePathWithTimestamp` to prevent confusion with `IndexedResource` and `IndexEntry`. Do the names make things clearer?

I also ran into a bug where the root path prefix wasn't being stripped from the files in some cases. This issue has been fixed.


Now, all that's left is to fix the build. A couple of tests are failing only on windows for some reason. I tried to resolve it today but couldn't get it to work. I still need to spend some time on it, because I rely on CI logs for debugging.

fs-index/src/index.rs
```diff
@@ -1,1120 +1,491 @@
...
-                pathdiff::diff_paths(path.to_str().unwrap(), self.root.clone())
-                    .ok_or(ArklibError::Path(
-                        "Couldn't calculate path diff".into(),
-                    ))?;
+        let root_path = root_path.as_ref();
+        // Canonicalize the root path (only for unix-like systems)
+        // Note: On windows, canonicalization adds a prefix to the path
+        // which breaks the path comparison in the tests
+        #[cfg(target_family = "unix")]
+        let root_path = root_path.canonicalize()?;
```
kirillt, line 289: Do we simply skip canonicalization on Windows now?

tareknaser: Please ignore this commit—it was just for troubleshooting. I’ll revert it in the next one. 
I didn’t want to make your life harder with force-pushing 😃

tareknaser:
Windows CI error fixed in e0f6e75

kirillt (approved):
Looks great!

PR #81, opened 2024-07-18
`fs-storage` : Add FolderStorage struct

Pushkarm029:
FileStorage struct retains core functionality from [Kotlin implementation](https://github.com/ARK-Builders/arklib-android/blob/main/lib/src/main/java/dev/arkbuilders/arklib/data/storage/FolderStorage.kt). https://github.com/ARK-Builders/ark-core/issues/1


fs-storage/src/folder_storage.rs
```diff
@@ -0,0 +1,336 @@
...
+where
+    K: Ord,
+{
+    /// Label for logging
+    label: String,
+    /// Path to the underlying file where data is persisted
+    path: PathBuf,
+    /// Last modified time of internal mapping. This becomes equal to
+    /// `written_to_disk` only when data is written or read from disk.
+    // modified: SystemTime,
```
twitu, line 38: Remove copied variable names

fs-storage/src/folder_storage.rs
```diff
@@ -0,0 +1,336 @@
...
+const STORAGE_VERSION: i32 = 3;
+
+/// Represents a file storage system that persists data to disk.
+pub struct FolderStorage<K, V>
+where
+    K: Ord,
+{
+    /// Label for logging
+    label: String,
+    /// Path to the underlying file where data is persisted
```
twitu, line 34: This documentation is mostly taken from FileStorage. While it is similar it should be updated to reflect FolderStorage specific details -
* Change all instances of FileStorage to FolderStorage
* `path` is the path to the directory not the file
* the key is the path of the file inside the directory
* The ram and disk timestamps are stored for each individual file

fs-storage/src/folder_storage.rs
```diff
@@ -0,0 +1,336 @@
...
+    path: PathBuf,
+    /// Last modified time of internal mapping. This becomes equal to
+    /// `written_to_disk` only when data is written or read from disk.
+    // modified: SystemTime,
+    ram_timestamps: BTreeMap<K, SystemTime>,
```
twitu, line 39: We have three btrees here mapping the same key to different values. Have you considered merging these into a single tree that stores a single entry that contains the value, the ram timestamp and the disk timestamp? Investigate if this is a better approach.

fs-storage/src/folder_storage.rs
```diff
@@ -0,0 +1,336 @@
...
+            let new_timestamp = SystemTime::now();
+            file.set_modified(new_timestamp)?;
+            file.sync_all()?;
+
+            self.disk_timestamps
+                .insert(key.clone(), new_timestamp);
+        }
+
+        // Remove files for keys that no longer exist
+        for entry in fs::read_dir(&self.path)? {
```
twitu, line 270: Extract this "clean up" logic into a separate function. It might be useful in other places too.

fs-storage/src/folder_storage.rs
```diff
@@ -0,0 +1,336 @@
...
+            }
+        }
+    }
+
+    /// Read the data from file
+    fn read_fs(&mut self) -> Result<&BTreeMap<K, V>> {
+        let data = self.load_fs_data()?;
+        self.data = data;
+        self.disk_timestamps.clear();
+        for key in self.data.entries.keys() {
```
twitu, line 225: Instead of collecting the disk timestamps separately. Why not collect them inside `load_fs_data` itself? Where you are already reading the data from the files you can also read the metadata.

twitu (changes requested):

fs-storage/src/folder_storage.rs
Pushkarm029: I tried it, but it makes the implementation less readable.

fs-storage/src/folder_storage.rs
```diff
@@ -195,11 +206,73 @@ where
...
+                    ram_newer = true;
+                }
+            } else {
+                // If the file doesn't exist on disk, RAM is newer
+                ram_newer = true;
+            }
+
+            // If we've found both RAM and disk modifications, we can stop checking
+            if ram_newer && disk_newer {
+                break;
```
twitu, line 237: Checking for files on disk that aren't in ram is not needed if this divergent condition has already been reached.

fs-storage/src/folder_storage.rs
```diff
@@ -195,11 +206,73 @@ where
...
+            if let Ok(metadata) = fs::metadata(&file_path) {
+                if let Ok(disk_timestamp) = metadata.modified() {
+                    match ram_timestamp.cmp(&disk_timestamp) {
+                        std::cmp::Ordering::Greater => ram_newer = true,
+                        std::cmp::Ordering::Less => disk_newer = true,
+                        std::cmp::Ordering::Equal => {}
+                    }
+                } else {
+                    // If we can't read the disk timestamp, assume RAM is newer
+                    ram_newer = true;
```
twitu, line 228: Adding some `log::debug` statements where the a variable changes and which file caused it to change will help in debugging later in case tests/app fails.

fs-storage/src/utils.rs
```diff
@@ -52,6 +54,38 @@ where
     Ok(data)
 }
 
+pub fn remove_files_not_in_ram<K, V>(
```
twitu, line 57: This "clean up" function is only relevant to `folder_storage` right now and can be kept in that module itself. Later if it becomes relevant to other storages it can be moved here.

Pushkarm029:
Tests are failing due to this:
```bash
error: unexpected `cfg` condition value: `unix`
 --> fs-atomic-versions/src/atomic/file.rs:3:7
  |
3 | #[cfg(target_os = "unix")]
  |       ^^^^^^^^^^^^^^^^^^
  |
  = note: expected values for `target_os` are: `aix`, `android`, `cuda`, `dragonfly`, `emscripten`, `espidf`, `freebsd`, `fuchsia`, `haiku`, `hermit`, `horizon`, `hurd`, `illumos`, `ios`, `l4re`, `linux`, `macos`, `netbsd`, `none`, `nto`, `openbsd`, `psp`, `redox`, `solaris`, `solid_asp3`, `teeos`, `tvos`, `uefi`, `unknown`, `visionos`, `vita`, `vxworks`, `wasi`, `watchos`, and `windows` and 2 more
  = note: see <https://doc.rust-lang.org/nightly/rustc/check-cfg/cargo-specifics.html> for more information about checking conditional configuration
  = note: `-D unexpected-cfgs` implied by `-D warnings`
  = help: to override `-D warnings` add `#[allow(unexpected_cfgs)]`

```

twitu (approved):
The CI seems to failing on some formatting/linting issues. Solid v1 implementation looks good to merge once that's fixed.

fs-storage/src/folder_storage.rs
```diff
@@ -0,0 +1,573 @@
...
+
+In version 2, `FolderStorage` stored data in a plaintext format.
+Starting from version 3, data is stored in JSON format.
+
+For backward compatibility, we provide a helper function `read_version_2_fs` to read version 2 format.
+*/
+const STORAGE_VERSION: i32 = 3;
+
+/// Represents a folder storage system that persists data to disk.
+pub struct FolderStorage<K, V>
```
tareknaser, line 30: In `FileStorage`, how we store the data was straightforward because we know that data is stored as a file with key-value pairs. But for `FolderStorage`, it's not immediately clear how the data is stored. We should add a doc comment to explain it. 

For example, if we're using `bincode` to serialize and deserialize the data where the key is the file name, we should mention that in a comment.

fs-storage/src/folder_storage.rs
```diff
@@ -0,0 +1,573 @@
...
+
+`FolderStorage` is a basic key-value storage system that persists data to disk.
+where the key is the path of the file inside the directory.
+
+
+In version 2, `FolderStorage` stored data in a plaintext format.
+Starting from version 3, data is stored in JSON format.
+
+For backward compatibility, we provide a helper function `read_version_2_fs` to read version 2 format.
+*/
```
tareknaser, line 26: This versioning information is specific to `FileStorage`. It is unrelated to `FolderStorage`.

Since this is our first iteration of `FolderStorage`, we should begin with version 1.

fs-storage/src/folder_storage.rs
tareknaser, line 1: Could we add an example for `FolderStorage` in `fs-storage/examples/cli.rs`? It could be as simple as a CLI argument to choose between file and folder storage, or a separate file if you prefer. 

I want to test `FolderStorage` on my file system.

fs-storage/src/folder_storage.rs
```diff
@@ -0,0 +1,573 @@
...
+where
+    K: Ord,
+{
+    /// Label for logging
+    label: String,
+    /// Path to the underlying file where data is persisted
+    path: PathBuf,
+    /// `ram_timestamps` can be used to track the last time a file was modified in memory.
+    /// where the key is the path of the file inside the directory.
+    ram_timestamps: BTreeMap<K, SystemTime>,
```
tareknaser, line 40: In `FileStorage`, we used two concepts:
1. `modified`: The last time the internal mapping was changed.
2. `written_to_disk`: The last time data was saved to disk.

Do we need to define `ram_timestamps`? Is there a difference between `ram_timestamps` and the internal mapping timestamp in `FileStorage`? If not, we can stick with the existing convention from `FileStorage`.

Adding new terms might be confusing.

fs-storage/src/folder_storage.rs
```diff
@@ -0,0 +1,573 @@
...
+///
+/// This is the data that is serialized and deserialized to and from disk.
+#[derive(Serialize, Deserialize)]
+pub struct FolderStorageData<K, V>
+where
+    K: Ord,
+{
+    version: i32,
+    entries: BTreeMap<K, V>,
+}
```
tareknaser, line 58: The purpose of this struct in `FileStorage` was to create a data structure that could be directly serialized and deserialized to and from disk (currently as a JSON). However, in the case of `FolderStorage`, we’re storing data in files where the key is the filename and the value is the file’s content.

Given this approach, I don't see a need for this struct. It might be better to serialize and deserialize the data directly to and from the file. We can remove this struct and handle the serialization and deserialization directly.

fs-storage/src/folder_storage.rs
```diff
@@ -0,0 +1,573 @@
...
+Starting from version 3, data is stored in JSON format.
+
+For backward compatibility, we provide a helper function `read_version_2_fs` to read version 2 format.
+*/
+const STORAGE_VERSION: i32 = 3;
+
+/// Represents a folder storage system that persists data to disk.
+pub struct FolderStorage<K, V>
+where
+    K: Ord,
```
tareknaser, line 32: Since the key in `FolderStorage` is the filename, we should add a constraint that `K` must implement `std::fmt::Display`, right?

fs-storage/src/folder_storage.rs
```diff
@@ -0,0 +1,573 @@
...
+            let entry = entry?;
+            let path = entry.path();
+            if path.is_file()
+                && path.extension().map_or(false, |ext| ext == "bin")
+            {
+                let key = path
+                    .file_stem()
+                    .unwrap()
+                    .to_str()
+                    .unwrap()
```
tareknaser, line 143: So we can be sure this doesn't panic

fs-storage/src/folder_storage.rs
```diff
@@ -0,0 +1,573 @@
...
+/// A struct that represents the data stored in a [`FolderStorage`] instance.
+///
+///
+/// This is the data that is serialized and deserialized to and from disk.
+#[derive(Serialize, Deserialize)]
+pub struct FolderStorageData<K, V>
+where
+    K: Ord,
+{
+    version: i32,
```
tareknaser, line 56: Where should we store the version information? In `FileStorage`, it is serialized in a separate field. For `FolderStorage`, we need a way to track the version to ensure backward compatibility in the future. I have two ideas:

- Embed the storage version in each file's content as a separate field.
- Use a special file in the folder to specify the version.

fs-storage/src/folder_storage.rs
```diff
@@ -0,0 +1,573 @@
...
+
+        let mut data = FolderStorageData {
+            version: STORAGE_VERSION,
+            entries: BTreeMap::new(),
+        };
+
+        self.disk_timestamps.clear();
+        self.ram_timestamps.clear();
+
+        // read_version_2_fs : unimplemented!()
```
tareknaser, line 131: There is no version 2 of `FolderStorage`

fs-storage/src/folder_storage.rs
```diff
@@ -0,0 +1,573 @@
...
+                            self.label.clone(),
+                            "Failed to parse key from filename".to_owned(),
+                        )
+                    })?;
+
+                let mut file = File::open(&path)?;
+                let mut buffer = Vec::new();
+                file.read_to_end(&mut buffer)?;
+
+                let value: V = bincode::deserialize(&buffer).map_err(|e| {
```
tareknaser, line 156: Does each file contain one value?

fs-storage/src/folder_storage.rs
```diff
@@ -0,0 +1,573 @@
...
+                        self.disk_timestamps.insert(key.clone(), modified);
+                        self.ram_timestamps.insert(key, modified);
+                    }
+                }
+            }
+        }
+        Ok(data)
+    }
+
+    fn remove_files_not_in_ram(&mut self) {
```
tareknaser, line 175: We need improved error handling for `remove_files_not_in_ram()`. There are many places where this function could fail and cause a panic. It might be better to return a `Result<>` instead.

tareknaser, line 175: Sorry, I'm not clear on what this function does. It deserves a doc comment.

fs-storage/src/folder_storage.rs
```diff
@@ -0,0 +1,573 @@
...
+            file.sync_all()?;
+
+            self.disk_timestamps
+                .insert(key.clone(), new_timestamp);
+            self.ram_timestamps
+                .insert(key.clone(), new_timestamp);
+        }
+
+        // Remove files for keys that no longer exist
+        self.remove_files_not_in_ram();
```
tareknaser, line 393: Commenting out this line doesn’t break the code, and all tests pass as usual. 
Could you explain its purpose? Also, can we add tests to cover this line?

fs-storage/src/folder_storage.rs
```diff
@@ -0,0 +1,573 @@
...
+const STORAGE_VERSION: i32 = 3;
+
+/// Represents a folder storage system that persists data to disk.
+pub struct FolderStorage<K, V>
+where
+    K: Ord,
+{
+    /// Label for logging
+    label: String,
+    /// Path to the underlying file where data is persisted
```
tareknaser, line 36:
```suggestion
    /// Path to the underlying folder where data is persisted
```

fs-storage/src/folder_storage.rs
```diff
@@ -0,0 +1,573 @@
...
+
+        Ok(storage)
+    }
+
+    /// Load mapping from folder storage
+    fn load_fs_data(&mut self) -> Result<FolderStorageData<K, V>> {
+        if !self.path.exists() {
+            return Err(ArklibError::Storage(
+                self.label.clone(),
+                "File does not exist".to_owned(),
```
tareknaser, line 112:
```suggestion
                "Folder does not exist".to_owned(),
```

fs-storage/src/folder_storage.rs
```diff
@@ -0,0 +1,573 @@
...
+            ));
+        }
+
+        let mut data = FolderStorageData {
+            version: STORAGE_VERSION,
+            entries: BTreeMap::new(),
+        };
+
+        self.disk_timestamps.clear();
+        self.ram_timestamps.clear();
```
tareknaser, line 129: Why do we need to clear the timestamps? Could we add a comment to explain this? Also, adding a test for this would be helpful. All tests pass even without these two lines.

fs-storage/src/folder_storage.rs
```diff
@@ -0,0 +1,573 @@
...
+        self.data.entries.insert(key.clone(), value);
+        self.ram_timestamps.insert(key, SystemTime::now());
+    }
+
+    /// Remove an entry from the internal mapping given a key
+    fn remove(&mut self, id: &K) -> Result<()> {
+        self.data.entries.remove(id).ok_or_else(|| {
+            ArklibError::Storage(self.label.clone(), "Key not found".to_owned())
+        })?;
+        // self.ram_timestamps.remove(id);
```
tareknaser, line 228: Remove commented out code

fs-storage/src/folder_storage.rs
```diff
@@ -0,0 +1,573 @@
...
+use std::io::{Read, Write};
+use std::time::SystemTime;
+use std::{
+    collections::BTreeMap,
+    path::{Path, PathBuf},
+};
+
+use crate::base_storage::{BaseStorage, SyncStatus};
+use crate::monoid::Monoid;
+// use crate::utils::read_version_2_fs;
```
tareknaser, line 12: Remove commented out code

tareknaser (changes requested):

fs-storage/src/folder_storage.rs
kirillt: Good point. Technically, we could solve the task for arbitrary key type by storing the keys together with values. But this complicates the solution unnecessary for this moment. We'll be able to loose the constraint when we implement "chunked storage".

fs-storage/src/folder_storage.rs
kirillt: Files storing the values should not be used for this. Examples of values stored in a `FolderStorage` range from JSON properties to binary image data.

Storing the version in a separate file sounds like a better idea to me. But do we really need versioning here?

Pushkarm029:
Thanks for the review

fs-storage/src/folder_storage.rs
tareknaser: We might manage without versioning information in `FolderStorage`, but if we decide to update to a newer version later (for instance, if binary serialization turns out to be unsuitable), implementing backward compatibility will become more challenging, though not impossible. 

We've already dealt with a similar situation in `FileStorage`.
I don't have a strong opinion on this. If you prefer to omit versioning information, that's fine with me.

kirillt: Agree. We might change the folder structure in future, too. Let's add version file just for the case.

On other hand, versioning of file format probably should be delegated to user side, because we should just write bytes using writer that user passes to us. So, JSONs could be printed as text, or in binary form. Can we implement it in a way that user controls how they want to store JSONs?

fs-storage/src/folder_storage.rs
Pushkarm029: yes

fs-storage/src/folder_storage.rs
twitu: Now that I think of it deleting a file is a fast operation (compared to writing to it). It removes the file entry from the directory inode. We could potentially implement two delete functions -
* delete - the default operations delete entries from the mapping and the file on disk
* bulk_delete - the default operations

fs-storage/src/folder_storage.rs
twitu: ChunkedStorage is meant to store n key-value pairs per file. FolderStorage is a special case where n = 1. We can optimise the special case by storing the key only in the in-memory mapping. This means that the key only needs constraints to be stored in the BTreeMap which would be Ord + Eq.

fs-storage/src/folder_storage.rs
twitu: I agree timestamps should not be cleared. Any values that need to be changed will be overwritten by the below two lines. Moreover, it stores timestamps of entries that were deleted so clearing it will wipe out that information. 

```rust
                        self.disk_timestamps.insert(key.clone(), modified);
                        self.ram_timestamps.insert(key, modified);
```

fs-storage/src/folder_storage.rs
twitu: We can remove all versioning related information since we do need it right now since our one and default serde format is bincode (which is optimized for fast serde).

We can later add versioning information as a drop in by inserting a `.version` file into the folder with specific custom serde format. If it doesn't exist, the implementation can default to bincode.

kirillt: Yeah, we don't need versioning of value files here, because serialization is provided by the user.

fs-storage/src/folder_storage.rs
```diff
@@ -0,0 +1,598 @@
...
+    /// without overwriting it.
+    pub fn new(label: String, path: &Path) -> Result<Self> {
+        let mut storage = Self {
+            label,
+            path: PathBuf::from(path),
+            ram_timestamps: BTreeMap::new(),
+            disk_timestamps: BTreeMap::new(),
+            data: FolderStorageData {
+                entries: BTreeMap::new(),
+            },
```
kirillt, line 92: Can we just have `entries` field directly without additional `FolderStorageData` structure?

fs-storage/src/folder_storage.rs
kirillt: Yes, but the easiest way to persist the keys is in the filenames.

```
<storage_root>
|- qwe
|- asd
|- zxc
```

`"qwe"`, `"asd"`, `"zxc"` are the keys here, e.g. file `<storage_root>/qwe` contains value assigned to key `"qwe"`

Do we store data like this currently? We could just fix the keys type to `String` but we want to use the storage at least with numerical keys, too. So, `Display` is the most practical generalization.

If we allow keys to be just `Ord`, then we need to serialize, store, load and deserialize them. I would postpone such work till we start working on the chunked storage.

fs-storage/src/folder_storage.rs
```diff
@@ -0,0 +1,598 @@
...
+
+    /// Get a value from the internal mapping
+    fn get(&self, id: &K) -> Option<&V> {
+        self.data.entries.get(id)
+    }
+
+    /// Write the data to folder
+    ///
+    /// Update the modified timestamp in file metadata to avoid OS timing issues
+    /// https://github.com/ARK-Builders/ark-rust/pull/63#issuecomment-2163882227
```
kirillt, line 390: We should not have links to discussions in the code, better to just explain the idea in simple words.

fs-storage/src/folder_storage.rs
kirillt: I'm not sure why do we need `bincode`. Will users be able to decode the value files themselves using simple UNIX tools? If `bincode` creates some optimized layout, then users would be required to use it for deserializing, too.

We should give users choice how they want to serialize their values. For text values though, UTF-8 is kinda standard encoding. Let's just bound the value types with `Serialize` trait.

Test code can use `bincode` or whatever is easier, probably just UTF-8 encoding for String values.

fs-storage/src/folder_storage.rs
```diff
@@ -0,0 +1,598 @@
...
+        Ok(data)
+    }
+
+    fn remove_files_not_in_ram(&mut self) -> Result<()> {
+        let dir = fs::read_dir(&self.path).map_err(|e| {
+            ArklibError::Storage(
+                self.label.clone(),
+                format!("Failed to read directory: {}", e),
+            )
+        })?;
```
kirillt, line 183: We should avoid re-scanning the storage directory if possible.

The only use-case for re-scanning seems to be disk corruption, or user messing with it manually. Let's do something simpler for the first version and create the follow-up task for future work on protecting the storage from user mistake or disk corruption.

fs-storage/examples/cli.rs
kirillt, line 1: @kirillt try this

fs-storage/Cargo.toml
```diff
@@ -17,7 +17,7 @@ serde_json = "1.0.82"
 serde = { version = "1.0.138", features = ["derive"] }
 jni = { version = "0.21.1", optional = true }
 jnix = { version = "0.5.1", features = ["derive"], optional = true }
-
+bincode = "1.3"
```
kirillt, line 20: I'm not sure why do we need `bincode`. Will users be able to decode the value files themselves using simple UNIX tools? If `bincode` creates some optimized layout, then users would be required to use it for deserializing, too.

We should give users choice how they want to serialize their values. For text values though, UTF-8 is kinda standard encoding. Let's just bound the value types with `Serialize` trait.

Test code can use `bincode` or whatever is easier, probably just UTF-8 encoding for String values.

fs-storage/src/folder_storage.rs
```diff
@@ -0,0 +1,598 @@
...
+        self.ram_timestamps
+            .insert(id.clone(), SystemTime::now());
+        Ok(())
+    }
+
+    /// Compare the timestamp of the storage files
+    /// with the timestamps of the in-memory storage and the last written
+    /// to time to determine if either of the two requires syncing.
+    fn sync_status(&self) -> Result<SyncStatus> {
```
kirillt, line 262: @kirillt continue reviewing from here

fs-storage/src/folder_storage.rs
```diff
@@ -0,0 +1,598 @@
...
+                        self.ram_timestamps.insert(key, modified);
+                    }
+                }
+            }
+        }
+        Ok(data)
+    }
+
+    fn remove_files_not_in_ram(&mut self) -> Result<()> {
+        let dir = fs::read_dir(&self.path).map_err(|e| {
```
kirillt, line 178: Although we need to detect external changes, we should avoid re-scanning the storage directory.

Do we really need to re-scan the storage directory here? Why not just delete? If the value file doesn't exist, then it either was deleted on remote side, or there is storage corruption (probably caused by the user messing with the storage manually). In standard case, when entries are deleted on remote side, we should detect deletions and delete entries from the in-memory mapping. This way it should be possible to avoid re-scans.

kirillt (changes requested):
Thanks for the great work, but we should polish raw edges here

twitu:
After experimenting with various synchronization models for concurrent operations, the "StrongDelete" approach is coming out to be the most effective in terms of code complexity and efficiency with passing tests.

The synchronization model can be stated as below -

    For concurrent insert and remove, the last writer wins
    For concurrent insert and external modification to disk, use monoidal combine to merge changes
    For concurrent delete and external modifications to disk, delete wins

fs-storage/src/folder_storage.rs
twitu: The proper way to allow the user to pick the serialization format would be to introduce another generic parameter and allow it to pick from a given format, or by making the generic parameter an implementation of Writer. However, I believe this can be a feature for later.

Currently, we can pick JSON as the default format since it is human readable. In fact, later when we extend the implementation for multiple writers for generics we don't need a separate version file because generic parameters can be set with default as JSON.

So for v1, I think not making it generic and using serde_json as the serialization format will be a good approach.

twitu (approved):

fs-storage/Cargo.toml
```diff
@@ -17,13 +17,15 @@ serde_json = "1.0.82"
...
-
 data-error = { path = "../data-error" }
 
 
 [dev-dependencies]
 anyhow = "1.0.81"
+quickcheck = { version = "1.0.3", features = ["use_logging"] }
+quickcheck_macros = "1.0.0"
 tempdir = "0.3.7"
+test-log = "0.2.16"
```
tareknaser, line 28: Not used

fs-storage/src/folder_storage.rs
```diff
@@ -0,0 +1,784 @@
...
+        for key in self.deleted_keys.iter() {
+            let file_path = self.path.join(format!("{}.json", key));
+            if file_path.exists() {
+                fs::remove_file(&file_path).expect("Failed to delete file");
+            }
+        }
+        Ok(())
+    }
+
+    pub fn extract_key_from_file_path(&self, path: &Path) -> Result<K> {
```
tareknaser, line 165: This method doesn't really depend on any fields or properties of `FolderStorage`, so it shouldn't be part of that struct. It would be better as a helper function, and it probably shouldn't be public.

fs-storage/src/folder_storage.rs
```diff
@@ -0,0 +1,784 @@
...
+    fn sync_status(&mut self) -> Result<SyncStatus> {
+        let mut ram_newer = !self.deleted_keys.is_empty();
+        let mut disk_newer = false;
+
+        for key in self.data.keys() {
+            let file_path = self.path.join(format!("{}.json", key));
+            let ram_timestamp = self
+                .modified
+                .get(key)
+                .expect("Data entry key should have ram timestamp");
```
tareknaser, line 233: Since we panic if the entry doesn't exist here,

fs-storage/src/folder_storage.rs
```diff
@@ -0,0 +1,784 @@
...
+                        ArklibError::Storage(
+                            self.label.clone(),
+                            err.to_string(),
+                        )
+                    })?;
+
+                data.insert(key.clone(), value);
+
+                if let Ok(metadata) = fs::metadata(&path) {
+                    if let Ok(modified) = metadata.modified() {
```
tareknaser, line 111: We should return an error here if we can't retrieve it from disk.

fs-storage/src/folder_storage.rs
```diff
@@ -0,0 +1,784 @@
...
+            if let Some(existing_value) = self.data.get(key) {
+                let existing_value_updated = self
+                    .modified
+                    .get(key)
+                    .and_then(|ram_stamp| {
+                        self.written_to_disk
+                            .get(key)
+                            .map(|disk_stamp| ram_stamp > disk_stamp)
+                    })
+                    .unwrap_or(false);
```
tareknaser, line 136: Why are we defaulting to `false` here?

fs-storage/src/folder_storage.rs
```diff
@@ -0,0 +1,784 @@
...
+                    }
+                }
+            }
+        }
+
+        self.data = data;
+        Ok(())
+    }
+    /// Resolve differences between memory and disk data
+    fn resolve_divergence(&mut self) -> Result<()> {
```
tareknaser, line 123: I'm sorry, but could you explain what this method does?

fs-storage/src/folder_storage.rs
```diff
@@ -0,0 +1,784 @@
...
+        };
+
+        log::info!("{} sync status is {}", self.label, new_status);
+        Ok(new_status)
+    }
+
+    /// Sync the in-memory storage with the storage on disk
+    fn sync(&mut self) -> Result<()> {
+        match self.sync_status()? {
+            SyncStatus::InSync => {}
```
tareknaser, line 318: We can log that the memory is synchronized with the storage.

fs-storage/src/folder_storage.rs
```diff
@@ -0,0 +1,784 @@
...
+            None => Err(ArklibError::Storage(
+                self.label.clone(),
+                "Key not found".to_owned(),
+            )),
+        }
+    }
+
+    /// Compare the timestamp of the storage files
+    /// with the timestamps of the in-memory storage and the last written
+    /// to time to determine if either of the two requires syncing.
```
tareknaser, line 223: Sorry, I don't fully understand this sentence.

fs-storage/src/folder_storage.rs
```diff
@@ -0,0 +1,784 @@
...
+                        std::cmp::Ordering::Less => {
+                            disk_newer = true;
+                            log::debug!(
+                                "Disk newer: file {} is newer on disk, ram: {}, disk: {}",
+                                file_path.display(),
+                                ram_timestamp.elapsed().unwrap().as_secs(),
+                                disk_timestamp.elapsed().unwrap().as_secs()
+                            );
+                        }
+                        std::cmp::Ordering::Equal => {}
```
tareknaser, line 254: Could we use `>` and `<` instead? I feel it's more intuitive.

fs-storage/src/folder_storage.rs
```diff
@@ -0,0 +1,784 @@
...
+        + serde::Serialize
+        + serde::de::DeserializeOwned
+        + std::str::FromStr
+        + std::fmt::Display,
+    V: Clone + serde::Serialize + serde::de::DeserializeOwned + Monoid<V>,
+{
+    /// Set a key-value pair in the internal mapping
+    fn set(&mut self, key: K, value: V) {
+        self.data.insert(key.clone(), value);
+        self.deleted_keys.remove(&key);
```
tareknaser, line 203: Just curious, why do we need to remove it from the deleted keys?

fs-storage/src/folder_storage.rs
```diff
@@ -0,0 +1,784 @@
...
+    K: Ord
+        + Clone
+        + serde::Serialize
+        + serde::de::DeserializeOwned
+        + std::str::FromStr
+        + std::fmt::Display,
+    V: Clone + serde::Serialize + serde::de::DeserializeOwned + Monoid<V>,
+{
+    /// Create a new folder storage with a diagnostic label and directory path
+    /// The storage will be initialized using the disk data, if the path exists
```
tareknaser, line 51: This is redundant since the same note is mentioned in the next line.

fs-storage/Cargo.toml
tareknaser, line 28: This crate isn't used anywhere.

fs-storage/README.md
```diff
@@ -42,5 +46,5 @@ key2: value2
 - To get all key value pairs
 
 ```bash
-cargo run --example cli read /tmp/z
+cargo run --example cli [file|folder] read /tmp/z
```
tareknaser, line 49: This command fails to print all key-value pairs when using folder storage.

fs-storage/examples/cli.rs
```diff
@@ -107,5 +125,87 @@ fn write_command(args: &[String], path: &str) -> Result<()> {
             }
         }
     }
+    fs.write_fs().expect("Failed to write to file");
+    Ok(())
+}
+
+fn folder_read_command(args: &[String], path: &str) -> Result<()> {
+    let keys = if args.len() > 3 {
```
tareknaser, line 133: I think this should get it working.
```suggestion
    let keys = if args.len() > 4 {
```

tareknaser (commented):
Thank you for the excellent work. I tested the CLI too, and it works perfectly as expected.

I still want to take a closer look at the tests though, especially the fancy property testing 😃

fs-storage/src/folder_storage.rs
```diff
@@ -0,0 +1,784 @@
...
+    /// `modified` can be used to track the last time a file was modified in memory.
+    /// where the key is the path of the file inside the directory.
+    modified: BTreeMap<K, SystemTime>,
+    /// `written_to_disk` can be used to track the last time a file written or read from disk.
+    /// where the key is the path of the file inside the directory.
+    written_to_disk: BTreeMap<K, SystemTime>,
+    data: BTreeMap<K, V>,
+    /// Temporary store for deleted keys until storage is synced
+    deleted_keys: BTreeSet<K>,
+}
```
tareknaser, line 29: To simplify, we could combine `modified` and `written_to_disk` into a single `BTreeMap<K, (SystemTime, SystemTime)>`, which we can call `timestamps`. This would avoid duplicating keys in memory.

We could even take it further and merge this with the `data` map into one struct, with a field like `deleted` to indicate whether an entry is deleted or not.

Since we don’t serialize directly but loop over the key-value pairs, I don’t think this should be a problem.

fs-storage/src/folder_storage.rs
Pushkarm029: I tried it, but Clippy suggests using `cmp` & `match` instead of this comparison_chain.

```bash
warning: `if` chain can be rewritten with `match`
   --> fs-storage/src/folder_storage.rs:268:21
    |
268 | /                     if ram_timestamp > &disk_timestamp {
269 | |                         ram_newer = true;
270 | |                         log::debug!(
271 | |                             "RAM newer: file {} is newer in RAM",
...   |
281 | |                         );
282 | |                     }
    | |_____________________^
    |
    = help: consider rewriting the `if` chain to use `cmp` and `match`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#comparison_chain
    = note: `#[warn(clippy::comparison_chain)]` on by default

```

tareknaser: Ok then it's fine. Not a big deal
Thank you

fs-storage/src/folder_storage.rs
Pushkarm029: `deleted_keys` is a `BTreeSet`, which just stores the key to be deleted. that's why we also clear it after sync. if it was BtreeMap, then we just compare it, just as we do with other `timestamps`.

fs-storage/src/folder_storage.rs
Pushkarm029: It is just an enhanced version of 'merge_from' for resolving divergence. The method considers timestamps to detect keys where we need to use monoids to resolve conflicts.


Added this comment here:
```rs
    /// Resolves discrepancies between in-memory data and disk data by combining or
    /// overwriting values based on which version is more recent, ensuring consistency.

```

fs-storage/src/folder_storage.rs
Pushkarm029: @twitu 

btw, defaulting to `true` also works

fs-storage/src/folder_storage.rs
Pushkarm029: updated it:
```rs
    /// Determines the synchronization status between RAM and disk.
    /// Compares modification timestamps of files in RAM and on disk,
    /// checking for newer versions in path.

```

fs-storage/src/folder_storage.rs
twitu: This is to ensure that deleted keys and keys in the mapping are a disjoint status. This helps us avoid iterating over keys and then filtering for non-deleted keys.

fs-storage/src/folder_storage.rs
twitu: Yeah, storing modified and written_to_disk timestamps in the same mapping seems like a better approach since it brings all the metadata together. However, I think it's better to keep the actual value in a separate map as it keeps metadata and actual data separate from each other.

I believe this will be helpful when creating bindings for JNI. If they were stored in a single struct and we don't want to expose internal metadata through Java interface (for encapsulation, perf reasons) then we'd have to "project" the necessary fields out of the struct making the solution more complex.

I think the timestamps and data mapping is a good middle ground.

fs-storage/src/folder_storage.rs
twitu: The `existing_value_updated` triggers a monoidal combine of values for a particular key.

Right now it works because the property based test uses a Dummy value which has a "dumb" monoidal implementation which does nothing i.e. combine(dummy, dummy) = dummy.

However, when we use values like strings or numbers that have useful monoidal implementations, we want to appropriately set the `existing_value_updated` flag. We want to combine things only if values have diverged in both memory and storage. So the default should be false, if the value is not updated it is simply set and not combined.

fs-storage/src/folder_storage.rs
kirillt: Interesting, maybe if it was one `if` after another

fs-storage/src/folder_storage.rs
kirillt: Am I right, that we keep keys in `deleted_keys` until they are propagated to other devices ("soft delete")? And if we encounter same key introduced again, then we remove it from "soft delete"?

(Just guessing, haven't read the code yet)

fs-storage/examples/cli.rs
kirillt: Doesn't `clap` handle such details automatically?

fs-storage/src/folder_storage.rs
twitu: Yup that is how it is used

fs-storage/examples/cli.rs
twitu: Currently, we are not using `clap` in the example

tareknaser (approved):
Thank you. No more comments from my side.

kirillt (approved):

PR #85, opened 2024-08-10
`fs-index`: Track API 

### LINKED ISSUE #84: `fs-index`: Track API
* extract Track API from [this PR](https://github.com/ARK-Builders/arklib/pull/72) and create it in this repo
* port comments to the new PR
    * https://github.com/ARK-Builders/ark-core/pull/79#discussion_r1665793019
    * https://github.com/ARK-Builders/ark-core/pull/79#discussion_r1665797041
    * https://github.com/ARK-Builders/ark-core/pull/79#discussion_r1665810017

tareknaser:
## Description

This PR adds a new method to the `ResourceIndex` in the `fs-index` crate. The added method, `update_one`, allows the index to be updated with the latest information from the file system for a single resource. It handles updates for resources that have been added, removed, or modified.

Fixes #84
## Changes
- **New Method**: `ResourceIndex::update_one`
- **Added Tests**:
	- `test_track_removal`
	- `test_track_modification`
	- `test_track_removal_with_collision`
	- `test_track_addition`
	- `test_track_modification_with_collision`
	- `test_track_addition_with_collision`
- **Added Benchmarks**:
	- `index_update_one`

### DIFF
--- a/fs-index/README.md
+++ b/fs-index/README.md
@@ -10,10 +10,11 @@ The most important struct in this crate is `ResourceIndex` which comes with:
 
 - **Reactive API**
   - `update_all`: Method to update the index by rescanning files and returning changes (additions/deletions).
-  - `update_one`: Method to update the index by rescanning a single file (addition/deletion/modification).
 - **Snapshot API**
   - `get_resources_by_id`: Query resources from the index by ID.
   - `get_resource_by_path`: Query a resource from the index by its path.
+- **Selective API**
+  - `update_one`: Method to manually update a specific resource by selectively rescanning a single file.
 
 ## Custom Serialization
 
--- a/fs-index/README.md
+++ b/fs-index/README.md
@@ -10,6 +10,7 @@ The most important struct in this crate is `ResourceIndex` which comes with:
 
 - **Reactive API**
   - `update_all`: Method to update the index by rescanning files and returning changes (additions/deletions).
+  - `update_one`: Method to update the index by rescanning a single file (addition/deletion/modification).
 - **Snapshot API**
   - `get_resources_by_id`: Query resources from the index by ID.
   - `get_resource_by_path`: Query a resource from the index by its path.
--- a/fs-index/benches/resource_index_benchmark.rs
+++ b/fs-index/benches/resource_index_benchmark.rs
@@ -112,24 +112,35 @@ fn resource_index_benchmark(c: &mut Criterion) {
                 ResourceIndex::build(black_box(&update_one_benchmarks_dir))
                     .unwrap();
 
-            // Create a new file
-            let new_file = update_one_benchmarks_dir.join("new_file.txt");
-            std::fs::File::create(&new_file).unwrap();
-            std::fs::write(&new_file, "Hello, World from new file!").unwrap();
+            // Create 1000 new files
+            for i in 5000..6000 {
+                let new_file =
+                    update_one_benchmarks_dir.join(format!("file_{}.txt", i));
+                std::fs::File::create(&new_file).unwrap();
+                std::fs::write(&new_file, format!("Hello, World! {}", i))
+                    .unwrap();
+            }
 
-            // Modify an existing file
-            let modified_file = update_one_benchmarks_dir.join("file_0.txt");
-            std::fs::write(&modified_file, "Hello, World from modified file!")
-                .unwrap();
+            // Modify 1000 files
+            for i in 4000..5000 {
+                let modified_file =
+                    update_one_benchmarks_dir.join(format!("file_{}.txt", i));
+                std::fs::write(&modified_file, format!("Bye, World! {}", i))
+                    .unwrap();
+            }
 
-            // Remove an existing file
-            let removed_file = update_one_benchmarks_dir.join("file_1.txt");
-            std::fs::remove_file(&removed_file).unwrap();
+            // Remove 1000 files
+            for i in 3000..4000 {
+                let removed_file =
+                    update_one_benchmarks_dir.join(format!("file_{}.txt", i));
+                std::fs::remove_file(&removed_file).unwrap();
+            }
 
-            // Update the index
-            let _update_result = index.update_one("new_file.txt").unwrap();
-            let _update_result = index.update_one("file_0.txt").unwrap();
-            let _update_result = index.update_one("file_1.txt").unwrap();
+            // Call update_one for each of the 3000 files
+            for i in 3000..6000 {
+                let file_path = format!("file_{}.txt", i);
+                let _update_result = index.update_one(&file_path).unwrap();
+            }
         });
     });
 
--- a/fs-index/benches/resource_index_benchmark.rs
+++ b/fs-index/benches/resource_index_benchmark.rs
@@ -84,6 +84,55 @@ fn resource_index_benchmark(c: &mut Criterion) {
         });
     });
 
+    // Benchmark `ResourceIndex::update_one()`
+
+    // First, create a new temp directory specifically for the update_one
+    // benchmark since we will be creating new files, removing files, and
+    // modifying files
+
+    let update_one_benchmarks_dir =
+        TempDir::with_prefix("ark-fs-index-benchmarks-update-one").unwrap();
+    let update_one_benchmarks_dir = update_one_benchmarks_dir.path();
+
+    group.bench_function("index_update_one", |b| {
+        b.iter(|| {
+            // Clear the directory
+            std::fs::remove_dir_all(&update_one_benchmarks_dir).unwrap();
+            std::fs::create_dir(&update_one_benchmarks_dir).unwrap();
+
+            // Create 5000 new files
+            for i in 0..5000 {
+                let new_file =
+                    update_one_benchmarks_dir.join(format!("file_{}.txt", i));
+                std::fs::File::create(&new_file).unwrap();
+                std::fs::write(&new_file, format!("Hello, World! {}", i))
+                    .unwrap();
+            }
+            let mut index: ResourceIndex<Crc32> =
+                ResourceIndex::build(black_box(&update_one_benchmarks_dir))
+                    .unwrap();
+
+            // Create a new file
+            let new_file = update_one_benchmarks_dir.join("new_file.txt");
+            std::fs::File::create(&new_file).unwrap();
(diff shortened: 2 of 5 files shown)


fs-index/src/index.rs
```diff
@@ -474,4 +474,109 @@ impl<Id: ResourceId> ResourceIndex<Id> {
...
+    /// Track the modification of a resource
+    ///
+    /// This method accepts a resource ID and returns `Result::Ok` if the
+    /// resource was successfully updated in the index.
+    ///
+    /// **Note**: The called must ensure that:
+    /// - The index is up-to-date with the file system except for the updated
+    ///  resource
+    /// - The resource is already in the index
+    pub fn track_modification(&mut self, _id: Id) -> Result<()> {
```
kirillt, line 553: Original motivation was here:
- https://github.com/ARK-Builders/arklib/issues/38

The use-case was updating the resource in one app, after its modification in another app.

But now I think, this function is not needed, because:
1. The editor app can have different flow, like opening a resource by one path, and saving the modified version to another path. Reactive API is much more robust here, especially with `notify`. We should ensure that `update_all` is efficient, too. We had a bug in there causing very slow re-indexing every time a single update was made. That's why the `track_update` appeared originally.
2. If the editor overwrites the resource by same path, then it still can use combination of `track_deletion` and `track_addition`.

fs-index/src/index.rs
```diff
@@ -474,4 +474,109 @@ impl<Id: ResourceId> ResourceIndex<Id> {
...
+    /// Track the addition of a new resource
+    ///
+    /// This method accepts a resource ID and returns `Result::Ok` if the
+    /// resource was successfully added to the index.
+    ///
+    /// **Note**: The called must ensure that:
+    /// - The index is up-to-date with the file system except for the new
+    ///   resource
+    /// - The resource is not already in the index
+    pub fn track_addition(&mut self, _id: Id) -> Result<()> {
```
kirillt, line 487: The use-case is an app that created resource, and it's very unlikely to have this resource indexed already (user created new image for instance). Then we can just add the resource to the index, skipping the whole index update.

So, the caller must ensure that:
1. The index is up-to-date except this single id.
2. The id hasn't been indexed before.

All `track` functions are intended to be used by app developers for optimized index manipulations, but they should understand the circumstances.

kirillt: We also can enforce invariant using `debug_assert!` that the id hasn't been indexed before.

fs-index/src/index.rs
```diff
@@ -474,4 +474,109 @@ impl<Id: ResourceId> ResourceIndex<Id> {
...
+    /// Track the removal of a resource
+    ///
+    /// This method accepts a resource ID and returns `Result::Ok` if the
+    /// resource was successfully removed from the index.
+    ///
+    /// **Note**: The called must ensure that:
+    /// - The index is up-to-date with the file system except for the removed
+    ///  resource
+    /// - The resource is already in the index
+    pub fn track_removal(&mut self, _id: Id) -> Result<()> {
```
kirillt, line 522: We can consider enforcing invariant using `debug_assert!` that the id is present in the index. If we do so, we also can remove `Result` and either succeed or panic if the invariant is violated.

fs-index/src/index.rs
kirillt: **tl;dr**: consider removing

fs-index/src/index.rs
kirillt: Now `update_one` makes sense to me again... 😅

fs-index/src/index.rs
kirillt: Consider merging with `update_one` when we pass the removed path and just remove it from the index together with the current id.

fs-index/src/index.rs
tareknaser: `update_one()` is quite similar to `track_modification()` in that both aim to update a resource, whether it was modified, added, or deleted. The main difference I see is that in this PR, `track_modification()` was designed to accept the resource ID instead of its path.

I support having a method like `update_one()` that checks the file system for changes and updates the index for just that specific resource. This would be useful in cases where the user doesn't want to run `update_all()`, which performs a full rescan.

The function definition should look something like this:
```rust
/// Update the index with the latest information from the file system
/// for a single resource
///
/// This method accepts the path of a single resource and updates the index
/// regardless of whether the resource was added, removed, or modified.
///
/// **Note**: The caller must ensure that:
/// - The index is up-to-date with the file system except for the updated
///  resource
/// - The resource was already in the index
pub fn update_one(&mut self, _path: &Path) -> Result<()>
```
We could also return a more informative type that provides details about the specific kind of update that occurred in the file system.

fs-index/src/index.rs
tareknaser: > We can consider enforcing invariant using `debug_assert!` that the id is present in the index.

If we decide to accept the path as a parameter in `track_addition()` as mentioned [here](https://github.com/ARK-Builders/ark-core/pull/85#discussion_r1718357077), it would be more consistent to also accept the path instead of the resource ID in `track_deletion()`.

That said, I believe [having a single function to update the index for one resource](https://github.com/ARK-Builders/ark-core/pull/85#discussion_r1719841869) might be more effective than separate functions for `track_addition()` and `track_deletion()`. What are your thoughts?

fs-index/README.md
```diff
@@ -10,6 +10,7 @@ The most important struct in this crate is `ResourceIndex` which comes with:
 
 - **Reactive API**
   - `update_all`: Method to update the index by rescanning files and returning changes (additions/deletions).
+  - `update_one`: Method to update the index by rescanning a single file (addition/deletion/modification).
```
kirillt, line 13: Should be mentioned in other section. "Reactive" means that we have a stream of updates and we react on these updates, even if we manually poll the stream. But this "cherry-pick" is completely manual approach, not reactive.

fs-index/benches/resource_index_benchmark.rs
```diff
@@ -84,6 +84,55 @@ fn resource_index_benchmark(c: &mut Criterion) {
...
+                .unwrap();
+
+            // Remove an existing file
+            let removed_file = update_one_benchmarks_dir.join("file_1.txt");
+            std::fs::remove_file(&removed_file).unwrap();
+
+            // Update the index
+            let _update_result = index.update_one("new_file.txt").unwrap();
+            let _update_result = index.update_one("file_0.txt").unwrap();
+            let _update_result = index.update_one("file_1.txt").unwrap();
```
kirillt, line 132: Right idea of the benchmark, but can we make more calls to `fn update_one`? Right now, there is too much of initialization each iteration, but the measured call takes too few. Maybe we should make the number of "new" or "modified" files to be comparable with number of "existing" files?

fs-index/src/index.rs
```diff
@@ -474,4 +470,82 @@ impl<Id: ResourceId> ResourceIndex<Id> {
...
+    /// **Note**: The caller must ensure that:
+    /// - The index is up-to-date with the file system except for the updated
+    ///   resource
+    /// - In case of a addition, the resource was not already in the index
+    /// - In case of a modification or removal, the resource was already in the
+    ///   index
+    pub fn update_one<P: AsRef<Path>>(
+        &mut self,
+        relative_path: P,
+    ) -> Result<()> {
```
kirillt, line 490: Nice. I think this API is the way to go since it's simple and clear.

kirillt: It should be considered experimental API anyway though, while "Reactive API" should be encouraged.

fs-index/src/index.rs
```diff
@@ -474,4 +470,82 @@ impl<Id: ResourceId> ResourceIndex<Id> {
...
+        // Check if the entry exists in the file system
+        if !entry_path.exists() {
+            // If the entry does not exist in the file system, it's a removal
+
+            // Remove the resource from the path to ID map
+            let id = self.path_to_id.remove(path).ok_or_else(|| {
+                ArklibError::Path(format!(
+                    "Path {:?} not found in the index",
+                    path
+                ))
```
kirillt, line 503: It's OK to replace the error message with `debug_assert!` since it's the assumption that must be ensured by the caller.

fs-index/src/tests.rs
```diff
@@ -666,3 +666,245 @@ fn test_update_all_files_with_same_hash() {
         assert_eq!(index.collisions().len(), 1, "{:?}", index);
     });
 }
+
+/// Simple test for tracking a single resource addition.
```
kirillt, line 670: In the `arklib` PR we also tried to implement an automated test scenario, where we'd test one specific invariant that "any sequence of `update_one` calls must be equivalent to `update_all` on the full data". My idea how to approach it is something like:
1. Start with empty directory.
2. Generate an update, randomly one of 3 kinds (modification and deletion could be generated only when we have some files). 
3. Call `update_one` on the updated file.
4. In isolation, build new index from the current state of the directory.
5. Compare outputs of step 3 and step 4. Must be equal.

This is not exactly a unit test, because it's slow and tests multiple corner cases in one scenario. But I think it must be useful to run it periodically. By the way, I think that Ishan and Pushkar do in `fs-storage` with property-based testing should reflect the same idea.

Do you think it's feasible to implement in this PR?

fs-index/README.md
tareknaser: Good point. What about the term "Selective API"?

fs-index/src/index.rs
tareknaser: Added a note for specifically this

fs-index/benches/resource_index_benchmark.rs
tareknaser: Agreed. Thanks for the suggestions. I’ve increased the number of `update_one()` calls to handle:
- 1000 file creations
- 1000 file modifications
- 1000 file removals

fs-index/src/tests.rs
tareknaser: Sounds like a plan.
We could create multiple tests to verify this as well. 
To do this properly and cleanly, we’ll likely need to set up some helper functions. 

Given the scope, I think this should be handled in its own PR. This current PR is blocking #36, so once it's merged, I can add the additional tests along with the index watch API. 

What do you think?

kirillt (approved):
Thanks, looks great 👍

PR #86, opened 2024-08-12
Add property based testing for FolderStorage

twitu:
The test checks whether the following invariants holds -
* The storage state is InSync when it is created.
* Setting or removing a key value pair makes the storage stale
   * While storage is stale an external modification happens the storage status diverges
* Externally modifying the disk when the storage is in sync makes the memory stale
   * If a key is set or removed when the memory  is stale the status diverges

Currently the test fails indicating some subtle bug in the implementation. Relaxing the properties or fixing the implementation are two possible ways to get the test passing. @Pushkarm029

### DIFF
--- a/fs-storage/src/folder_storage.rs
+++ b/fs-storage/src/folder_storage.rs
@@ -10,20 +10,6 @@ use crate::base_storage::{BaseStorage, SyncStatus};
 use crate::monoid::Monoid;
 use data_error::{ArklibError, Result};
 
-/*
-Note on `FolderStorage` Versioning:
-
-`FolderStorage` is a basic key-value storage system that persists data to disk.
-where the key is the path of the file inside the directory.
-
-
-In version 2, `FolderStorage` stored data in a plaintext format.
-Starting from version 3, data is stored in JSON format.
-
-For backward compatibility, we provide a helper function `read_version_2_fs` to read version 2 format.
-*/
-// const STORAGE_VERSION: i32 = 3;
-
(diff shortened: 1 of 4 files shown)


fs-storage/src/folder_storage.rs
```diff
@@ -595,4 +589,260 @@ mod tests {
...
+
+    use data_error::Result;
+    use quickcheck::{Arbitrary, Gen};
+    use std::collections::{BTreeMap, HashSet};
+    use std::time::SystemTime;
+
+    // Assuming FolderStorage, BaseStorage, SyncStatus, and other necessary types are in scope
+
+    #[derive(Clone, Debug)]
+    enum StorageOperation {
```
kirillt, line 601: Is it possible to re-use the enum from `FileStorage` or operations differ indeed?

kirillt (approved):
Cool stuff, I like that

fs-storage/src/folder_storage.rs
twitu: Unfortunately, these are different operations.

twitu:
After experimenting with various synchronization models for concurrent operations, the "StrongDelete" approach is coming out to be the most effective in terms of code complexity and efficiency with passing tests.

The synchronization model can be stated as below -
* For concurrent insert and remove, the last writer wins
* For concurrent insert and external modification to disk, the last writer wins
* Between deletes and external modifications to disk, delete wins

PR #87, opened 2024-08-13
Inital Blocking "RPC" Proof of Concept

alvinosh:


nonblock-rpc/android/app/src/main/jniLibs/arm64-v8a/librpc_core.so
kirillt, line 1: Binary files should not be committed. They make cloning the Git repo slower once merged into `main`, forever.

kirillt: (4 `.so` files)

kirillt: @alvinosh

nonblock-rpc/rust/Cargo.toml
tareknaser, line 1: The Rust crate should be located directly in `ark-rust/nonblock-rpc` not in `/rust`

tareknaser, line 1: We should also include it in the `ark-rust` `Cargo.toml` file so it's part of the cargo workspace.

nonblock-rpc/rust/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[workspace]
+
+members = [
+    "core",
+    "uniffi-bindgen",
+]
```
tareknaser, line 6: This setup creates nested Rust workspaces, [which is generally discouraged](https://www.reddit.com/r/rust/comments/1b1lgkw/nested_workspace_how_can_you_work_around_the/). 

Do you think there's a way to refactor this? Perhaps `uniffi-bindgen` could be included as a binary within the same crate.

nonblock-rpc/README.md
tareknaser, line 1: Thanks for the build guide! Could we also add some usage instructions and perhaps an example to show the crate in action?

nonblock-rpc/README.md
```diff
@@ -0,0 +1,30 @@
...
+### Go to Rust Directory
+
+```sh
+cd rust/
+```
+
+### Build the dylib
+
+```sh
+cargo build
```
tareknaser, line 17: Could we consider adding the `--release` flag as well?
```suggestion
cargo build --release
```

nonblock-rpc/README.md
```diff
@@ -0,0 +1,30 @@
...
+### Build the Android libraries in jniLibs
+
+```sh
+cargo ndk -o ../android/app/src/main/jniLibs --manifest-path ./Cargo.toml -t armeabi-v7a -t arm64-v8a -t x86 -t x86_64 build --release
+```
+
+### Create Kotlin bindings
+
+```sh
+cargo run --bin uniffi-bingen generate --library ./target/debug/rpc_core.dll --language kotlin --out-dir ../android/app/src/main/java/ark/rpc_core
```
tareknaser, line 29:
```suggestion
cargo run --bin uniffi-bingen generate --library ./target/release/rpc_core.dll --language kotlin --out-dir ../android/app/src/main/java/ark/rpc_core
```

nonblock-rpc/android/.gitignore
tareknaser, line 1: Could you add some tests for the generated Kotlin bindings?  [`iroh` has simple tests for its generated bindings](https://github.com/n0-computer/iroh-ffi/tree/main/kotlin)—maybe we could take some inspiration from there.

nonblock-rpc/rust/core/src/rpc.rs
tareknaser, line 1: Why is this file commented out?

nonblock-rpc/rust/Cargo.toml
tareknaser, line 1: Could you also run the formatter with [the rules defined in the project root](https://github.com/ARK-Builders/ark-core/blob/main/rustfmt.toml) and run Clippy to address any linting issues?

nonblock-rpc/rust/core/src/lib.rs
```diff
@@ -0,0 +1,37 @@
...
+
+
+#[uniffi::export]
+pub fn my_func(map: BTreeMap<u64, String>) -> String {
+    let mut result = 0;
+    for (key, value) in map {
+        result += value;
+    }
+    result.to_string()
+}
```
tareknaser, line 37: Sorry, but I'm not quite clear on why we're exposing `factorial()`, `sum()`, and `my_func()` in this crate. Are they intended as examples? If so, it might be better to move them to an `/examples` directory.

nonblock-rpc/rust/core/src/api.rs
```diff
@@ -0,0 +1,13 @@
...
+
+#[uniffi::export]
+pub fn call(path: String, data: Vec<String>) -> String {
+    let routes = &ROUTER.routes;
+
+    match routes.get(&path) {
+        Some(handler) => {
+            handler.call(data)
+        },
+        None => "Unknown function".to_string(),
```
tareknaser, line 11: We're currently suppressing errors here. This method should return `Result<String>`, where the error type is a `uniffi` error type instead.

nonblock-rpc/rust/core/src/router.rs
```diff
@@ -0,0 +1,95 @@
...
+        }
+    }
+}
+
+#[derive(Serialize)]
+pub struct Response<T> {
+    pub result: Option<T>,
+    pub error: Option<String>,
+    pub is_success: bool,
+}
```
tareknaser, line 34: The fields of `Response<T>` shouldn't be public.

nonblock-rpc/rust/core/src/router.rs
```diff
@@ -0,0 +1,95 @@
...
+        self.routes.insert(name.to_string(), Box::new(handler));
+        self
+    }
+
+    pub fn call(&self, name: &str, args: Vec<String>) -> String {
+        match self.routes.get(name) {
+            Some(handler) => handler.call(args),
+            None => "Unknown function".to_string(),
+        }
+    }
```
tareknaser, line 26: `Unknown function` should be wrapped in an error type. The `call()` method should return `Result<String>` instead.

nonblock-rpc/rust/core/src/router.rs
```diff
@@ -0,0 +1,95 @@
...
+}
+
+#[derive(Serialize)]
+pub struct Response<T> {
+    pub result: Option<T>,
+    pub error: Option<String>,
+    pub is_success: bool,
+}
+
+const CATASTROPHIC_ERROR: &str = "{\"result\": null, \"error\": \"CATASTROPHIC_ERROR: Failed to serialize response\", \"is_success\": false}";
```
tareknaser, line 36: ‘CATASTROPHIC_ERROR’ sounds pretty intense 😃 
Is there a specific reason for using such a strong term?

By the way, I think this might be something that could be serialized into JSON.

nonblock-rpc/rust/core/src/api.rs
```diff
@@ -0,0 +1,13 @@
+use crate::ROUTER;
+
+#[uniffi::export]
+pub fn call(path: String, data: Vec<String>) -> String {
```
tareknaser, line 4: Nit: I think the term 'path' might be a bit confusing here. Perhaps 'route' would be a better fit.

tareknaser (changes requested):

nonblock-rpc/rust/Cargo.toml
kirillt: Actually, we might need split the crate in 2 parts: one in this repo, one in `ark-android` repo.

On other hand, we have some bindings in this repo. @tareknaser what should be final structure of the crate?

kirillt: About the crate name, maybe `ark-rpc` or just `rpc`?

tareknaser: > what should be final structure of the crate?


I think we should keep the bindings and the code that generates them in the same repo, since they depend on each other. For example, the command to build the bindings is:
```sh
cargo run --bin uniffi-bingen generate --library ./target/debug/rpc_core.dll --language kotlin --out-dir ../android/app/src/main/java/ark/rpc_core
```

This relies on the Rust code being built in the current folder, so any changes to the code would affect the bindings. That's why they should stay in the same Git repo.

I see two options:
- Keep all the Rust code in `ark-core/nonblock-rpc` and the bindings in `ark-core/nonblock-rpc/android`.
- Move the crate and bindings to a new repo. The crate can still depend on other `ark-core` crates as external dependencies.

tareknaser: > About the crate name, maybe `ark-rpc` or just `rpc`?

I think `rpc`, `block-rpc`, or `non-block-rpc` would work well. 
All the crates in this repository are related to ARK, it doesn't need `ark` in the name.

tareknaser: > I don't remember why we decided to keep `java` folder here.

Because they are bindings to the Rust code in this repository. For example, [here](https://github.com/ARK-Builders/ark-core/blob/main/java/lib/build.gradle.kts#L35), we are using the compiled Rust code to build the bindings:
```

    // Set the JVM argument for the java.library.path (To import rust compiled library)
    val rustLibPath = projectDir.resolve("../../target/release").absolutePath

```

Keeping the bindings in the same crate allows us to treat everything as one bundle of code. For example, we want the CI to run Java bindings checks for each PR in this repo, and that wouldn’t be possible if the bindings were in a different repository (as far as I know).

Another approach would be to move the code that generates the bindings in Rust to a separate repository that depends on `ark-core`. For instance, files in the `fs-storage/src/jni` folder could be moved to their own repository, with an external dependency on `ark-core`. 
The benefit of this is better organization, as we would divide the repositories by language. However, the downside is that we would have to maintain them separately. Any changes to the API in `ark-core` would need to be manually propagated in the other repository.

alvinosh: I think RPC makes more sense since in the future we want this one crate to support both blocking and non-blocking requests

twitu:
@kirillt suggested I take a look at this for a review. I do not have full context on it's implementation, so please help me in my understanding. Having implemented something similar previously, it seems to be a "client-server" like architecture where kotlin code is the client and can send requests to the rust server which can send back responses. The benefit of this approach is that it greatly simplifies dealing with generics because request/responses are strings or vec of strings.

I have some points to offer to enhance the understand-ability and maintainability of this module.
* Perhaps renaming it to function registry or method dispatcher will be a better fit. Because it is NOT RPC. RPC implies that the procedures are being called in a different address space i.e. a different process. Here the client and server exist in the same process, so calling it RPC can create confusion later on.
* It would be more appropriate to rename `HandlerFunction` to `Callback` because that is what a callback is. A pointer to a function stored on the heap. This will also help differentiate it from `FunctionHandler` struct :smile: 
* There are two traits `Handler` and `HandlerFunction` which have the same methods but different type constraints. I'm not sure what the difference is between the two. Making it clear with some documentation comments will help.
* The struct that actually stores the callback in this case the `FunctionHandler` would be more appropriately named as `RequestHandler`, since it is a handler that is triggered by a request from the client and internally calls a callback.

These terms have standard meanings which will make it much easier to understand and follow the logic. Of course some documentation will also help.

I found the `impl_handler_function` very useful and it should be documented well because it will come in handy if the callbacks need to be extended for more arguments.

kirillt:
@twitu good writeup, thanks. Yes, your understanding about client-server architecture is correct, and indeed we aim at applying it to a probably simpler task, where both sides are located not only on same device but also in same process.

I like your idea for the naming, "dispatch" sounds like a better word. Regarding "callback" term: technically, you are correct but it also might be an oversimplification. I like the terminology taken from Web, with "requests" and "responses". Especially, because we're aiming on async handling as well. Although when we implement subscription pattern, such a terminology could stand in our way.

Could you please split this list into smaller comments and leave them on related lines, so it would be easier to discuss and address?

rpc/src/router.rs
```diff
@@ -0,0 +1,153 @@
...
+
+    pub fn error(error: String) -> Self {
+        Response {
+            result: None,
+            error: Some(error),
+            is_success: false,
+        }
+    }
+}
+pub trait HandlerFunction<Marker>: Send + Sync + 'static {
```
twitu, line 77: Renaming this to `Callback` will be more appropriate because it represents a function that will be boxed and put on the heap and called with some value.

Moreover, having a different name will reduce confusion between it and FunctionHandler.

rpc/src/router.rs
```diff
@@ -0,0 +1,153 @@
...
+        }
+    };
+}
+
+impl_handler_function!(T0, T1);
+impl_handler_function!(T0, T1, T2);
+impl_handler_function!(T0, T1, T2, T3);
+impl_handler_function!(T0, T1, T2, T3, T4);
+
+struct FunctionHandler<F, Marker> {
```
twitu, line 138: Renaming this to `RequestHandler` will be more appropriate since it will accept a request and pass it to a "callback". It will also help differentiate it from `HandlerFunction` which can renamed as `Callback`.

rpc/src/router.rs
```diff
@@ -0,0 +1,153 @@
...
+
+impl<F: HandlerFunction<Marker>, Marker> Handler
+    for FunctionHandler<F, Marker>
+{
+    fn call(&self, args: Vec<String>) -> String {
+        self.function.call(args)
+    }
+}
+
+pub trait Handler {
```
twitu, line 151: I don't understand what this is doing and how it is different from `HandlerFunction`.

rpc/src/router.rs
```diff
@@ -0,0 +1,153 @@
...
+use serde::{Deserialize, Serialize};
+
+pub struct Router {
+    pub routes: HashMap<String, Box<dyn Handler + 'static + Send + Sync>>,
+}
+
+impl Router {
+    pub fn new() -> Self {
+        Router {
+            routes: HashMap::new(),
```
twitu, line 12: Pub-sub pattern can be supported by adding a routing table here which is a HashMap that stores the topic and a Vec of callbacks - `HashMap<String, Vec<HandlerFunction>>`. And exposing `publish` and `subscribe`, `unsubscribe` that allow publishing data and modifying subscriptions.

rpc/src/router.rs
```diff
@@ -0,0 +1,153 @@
+use std::{collections::HashMap, marker::PhantomData};
+
+use serde::{Deserialize, Serialize};
+
+pub struct Router {
+    pub routes: HashMap<String, Box<dyn Handler + 'static + Send + Sync>>,
```
twitu, line 6: Consider running the Router in a separate thread and using a single-thread per-core runtime like tokio-uring. This will make the code and types significantly less complex - you will be able to remove 'static, Send and Sync while still using an async runtime.

https://emschwartz.me/async-rust-can-be-a-pleasure-to-work-with-without-send-sync-static/

rpc/src/router.rs
```diff
@@ -0,0 +1,153 @@
...
+                _ => Response::error(
+                    "Failed to deserialize arguments".to_string(),
+                ),
+            }
+        };
+        serde_json::to_string(&response).unwrap_or(CATASTROPHIC_ERROR.into())
+    }
+}
+
+macro_rules! impl_handler_function {
```
twitu, line 105: Some documentation for this will be very helpful.

twitu (commented):
I've left a few comments on naming and design. If this works well then it's possible that we don't need JNI bindings for individual storages. The storages can be registered to this dispatcher and the application and can send "requests" to the dispatcher and get back "responses". This will also eliminate the problem of exposing generic bindings.

Is the plan to replace JNI bindings with this style?

nonblock-rpc/rust/Cargo.toml
twitu: Perhaps a different name like dispatcher or function registry will be more appropriate since this a single process method dispatch. The same name can be used for both sync and async.

nonblock-rpc/README.md
twitu: Thanks for including the full integration test with Kotlin. Having a couple of pure Rust tests that use the Router will also be very helpful. They will also be useful in quickly testing out changes. One other thing to consider is to include tests that call `HandlerFunctions` with different number of arguments like 3-4.

PR #92, opened 2024-10-24
Publish Java package

oluiscabral:
This action replicates the release functionality from `arklib-android` to GitHub Packages. Please test and confirm if it works as expected @kirillt

PR #94, opened 2024-11-01
Build `fs-storage` and publish Java package

oluiscabral:


.github/workflows/build.yml
```diff
@@ -55,6 +55,32 @@ jobs:
         run: gradle test
         working-directory: ./java
 
+      # Upload fs-storage JNI libs
+      - name: Install cargo-ndk
+        run: cargo install cargo-ndk
```
Pushkarm029, line 60: I suggest creating a separate `android` job instead of including these steps in the `linux` job. This separation would reduce execution time and simplify the workflow structure. 

The android-specific build could run in parallel with other platform builds.

.github/workflows/build.yml
```diff
@@ -55,6 +55,32 @@ jobs:
...
+          targets: aarch64-linux-android armv7-linux-androideabi i686-linux-android x86_64-linux-android
+
+      - name: Build fs-storage JNI libs
+        run: cargo ndk -o ./target/release/fs-storage/jniLibs --target aarch64-linux-android --target armv7-linux-androideabi --target i686-linux-android --target x86_64-linux-android build -p fs-storage --release
+
+      - name: Upload fs-storage JNI libs
+        uses: actions/upload-artifact@v4
+        with:
+          name: fs-storage-jni-libs
+          path: ./target/release/fs-storage/jniLibs
```
Pushkarm029, line 75: Do we need to release artifacts with every commit to `main`? I'm not sure.

.github/workflows/build.yml
```diff
@@ -55,6 +55,32 @@ jobs:
         run: gradle test
         working-directory: ./java
 
+      # Upload fs-storage JNI libs
+      - name: Install cargo-ndk
+        run: cargo install cargo-ndk
+
+      - name: Add Rust Toolchain
```
Pushkarm029, line 62: We can also consider adding cache-support to reduce workflow execution time:

```yaml
      - name: Setup Cargo Cache
        uses: Swatinem/rust-cache@v2
        with:
          save-if: false
```

java/lib/build.gradle.kts
```diff
@@ -29,14 +25,31 @@ java {
...
     options.encoding = "UTF-8"
     options.memberLevel = JavadocMemberLevel.PUBLIC
 }
+
+publishing {
+    // Define a Maven publication for the 'maven' repository
+    publications {
+        create<MavenPublication>("Maven") {
+            from(components["java"])
+        }
```
Pushkarm029, line 43: We can also add some necessary metadata here:

```kotlin
        create<MavenPublication>("mavenJava") {
            pom {
                name = "My Library"
                description = "A concise description of my library"
                url = "http://www.example.com/library"
                properties = mapOf(
                    "myProp" to "value",
                    "prop.with.dots" to "anotherValue"
                )
                licenses {
                    license {
                        name = "The Apache License, Version 2.0"
                        url = "http://www.apache.org/licenses/LICENSE-2.0.txt"
                    }
                }
                developers {
                    developer {
                        id = "johnd"
                        name = "John Doe"
                        email = "john.doe@example.com"
                    }
                }
                scm {
                    connection = "scm:git:git://example.com/my-library.git"
                    developerConnection = "scm:git:ssh://example.com/my-library.git"
                    url = "http://example.com/my-library/"
                }
            }
        }

```

Pushkarm029 (changes requested):
A few changes needed; overall, it looks good.

.github/workflows/build.yml
oluiscabral: That's what I did on the other Pull Request ([previous Pull Request](https://github.com/ARK-Builders/ark-core/pull/92#discussion_r1822047234)), but @kirillt sugestted to move all the setup to this one - `build.yml`, instead. While I still think moving it to a separate job is preferable, I'm interested in hearing @kirillt  reasoning for keeping it here.

.github/workflows/build.yml
oluiscabral: It seems like running the workflow with every commit is the current preference. I had suggested a more targeted approach based on specific project updates, but received feedback that continuous execution might be better.  I'm curious to hear everyone's thoughts on finding the most optimal solution.

.github/workflows/build.yml
oluiscabral: That's what I kind of did on the other Pull Request ([previous Pull Request](https://github.com/ARK-Builders/ark-core/pull/92#discussion_r1822047234)). While I still think moving it to a separate job is preferable, I'm interested in hearing what @kirillt thinks about it.

.github/workflows/build.yml
oluiscabral: You mean adding it to all the other jobs too?

java/lib/build.gradle.kts
oluiscabral: Interesting! Seems a bit overkill, but thanks for pointing that out. I'll include at least a name and description to it

.github/workflows/build.yml
Pushkarm029: Yes

.github/workflows/build.yml
Pushkarm029: I think @kirillt means to create a job in the same workflow, something like this:
```yaml
jobs:
  android:
    - uses: actions/checkout@v4
    - name: Add Rust Toolchain for android
    - name: Install cargo-ndk
    - name: Install JDK
    - name: Set up Gradle
    - name: Java tests
    - name: Build fs-storage JNI libs
    - name: Upload fs-storage JNI libs
    - name: Publish Java release

  linux:
    steps:
      - uses: actions/checkout@v4
      - name: Install Rust
      - name: Check
      - name: Format
      - name: Build Debug
      - name: Run tests
      - name: Build Release

  windows:
    steps:
      - uses: actions/checkout@v4
      - name: Install Rust
      - name: Build Release
      - name: Run tests

  mac-intel:
    steps:
      - uses: actions/checkout@v4
      - name: Install Rust
      - name: Build Release
      - name: Run tests

```

kirillt: In the past PR I actually was meaning one job for all-Java-related things because we had Java-related things scattered by 2 files (workflows).

Indeed, `android` job sounds like a good idea, but:
- initially, I thought that we keep `java` folder here with the idea of platform-agnostic bindings, i.e. same Java bindings could be used on desktop
- if this is not the case, then we should move the bindings to `ark-android` repo
- but given that Java bindings is temporary, we don't want to do redundant work, we even can wire `ark-android` to this branch of `ark-core` without merging into `main` (if it's possible)
- final goal is to have clean separation of Rust core which is unversal and multi-platform, from `ark-android` or other platform-dependent repos
- is UniFFI is quick to implement, we can remove `java` folder

kirillt: So, Java job should be separate and isolated from Rust jobs. Android job should be considered temporary because Core is multi-platform.

.github/workflows/build.yml
kirillt: Just uploading artifacts on every commit to `main` is fine and convenient for development, because we can take the lib and put it manually in ark-android. Artifacts should not be treated as releases, which we don't have set up in this repo. Artifacts expire after 3 month.

We need to setup proper releases as well, usually such are triggered by tagging a commit.

java/lib/build.gradle.kts
```diff
@@ -1,24 +1,20 @@
 plugins {
-    // Apply the java-library plugin for API and implementation separation.
-    `java-library`
+    `maven-publish` // Apply the maven-publish plugin before java-library for publishing to GitHub Packages.
```
kirillt, line 2: Will it be used together with workflow like [this one in the archived repo](https://github.com/ARK-Builders/arklib-android/blob/main/.github/workflows/release.yml)?

oluiscabral: Exactly. It allows the publish action:
```yml
    - name: Publish Release to Github
      uses: gradle/gradle-build-action@v2
      with:
        arguments: publishReleasePublicationToGithubPackagesRepository
      env:
        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

data-link/Cargo.toml
```diff
@@ -13,7 +13,7 @@ log = { version = "0.4.17", features = ["release_max_level_off"] }
 serde_json = "1.0.82"
 serde = { version = "1.0.138", features = ["derive"] }
 url = { version = "2.2.2", features = ["serde"] }
-reqwest = "0.11.11"
+reqwest = { version = "0.11.11", features = [ "json", "rustls-tls" ], default-features = false }
```
oluiscabral, line 16: @kirillt, this makes `OpenSSL` not required when compiling/building the project.

It's related to [Can't cross-compile project with OpenSSL](https://users.rust-lang.org/t/cant-cross-compile-project-with-openssl/70922/3)

data-error/Cargo.toml
```diff
@@ -10,7 +10,7 @@ bench = false
 
 [dependencies]
 thiserror = "1"
-reqwest = "0.11.11"
+reqwest = { version = "0.11.11", features = [ "json", "rustls-tls" ], default-features = false }
```
oluiscabral, line 13: @kirillt, this makes `OpenSSL` not required when compiling/building the project.

It's related to [Can't cross-compile project with OpenSSL](https://users.rust-lang.org/t/cant-cross-compile-project-with-openssl/70922/3)

ark-cli/README.md
```diff
@@ -1,32 +1,5 @@
...
-# Debian and Ubuntu
-$ sudo apt-get install pkg-config libssl-dev
-
-# Fedora
-$ sudo dnf install pkg-config perl-FindBin openssl-devel
-
-# Alpine Linux
-$ apk add pkgconfig openssl-dev
-```
-
```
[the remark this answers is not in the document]
oluiscabral: Actually, it is not required anymore. This was included just because `reqwest` requires that the machine compiling the project have `OpenSSL` installed, but this becomes tricky when it is intended to be cross-compiled.

Related to [Can't cross-compile project with OpenSSL](https://users.rust-lang.org/t/cant-cross-compile-project-with-openssl/70922/3)

java/lib/build.gradle.kts
oluiscabral: Done!

.github/workflows/build.yml
oluiscabral: I just added it

ark-cli/README.md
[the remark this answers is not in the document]
kirillt: Awesome, thanks!

.github/workflows/build.yml
Pushkarm029: No. All current jobs also build and test Rust code on each specific platform, which is important. 

Linux and Android will not be the same because the Linux job includes Rust tasks like `cargo build`, `cargo test` etc., while the Android job includes `cargo-ndk` build and other Android-specific tasks.

oluiscabral: Yes, @Pushkarm029, but maybe the missing point is: the Java tests and release depends on the built Rust code.

The Java sub-project is just JNI bindings for `fs-storage` and, in the future, if not replaced by `uniffi-rs`, it will also contain bindings for other crates 😬 So building and testing the Rust code is basically a "must have" too. The only extra thing compared to the Linux job would be the `cargo-ndk` related code

Pushkarm029: Oh, yeah. I forgot about that. Thanks for explaining it!
> The only extra thing compared to the Linux job would be the cargo-ndk related code.


Yes, I think to simplify everything, we have to accept this. So, we would have something like:
```yml
linux: build test
android: build test cargo-ndk
mac: build test
windows: build test
```

.github/workflows/build.yml
```diff
@@ -22,8 +22,14 @@ jobs:
...
         with:
-          toolchain: nightly # nightly is required for fmt
+          toolchain: nightly
           components: rustfmt, clippy
+          targets: aarch64-linux-android armv7-linux-androideabi i686-linux-android x86_64-linux-android
+
+      - name: Set up Cargo Cache
+        uses: Swatinem/rust-cache@v2
+        with:
+          save-if: false
```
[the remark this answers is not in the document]
kirillt: @Pushkarm029 I think that's a great idea. Just curious, where the cache will be really stored in this case?

@oluiscabral here is our weekly job, could you enable cache storage in it?
https://github.com/ARK-Builders/ark-core/actions/runs/11991530925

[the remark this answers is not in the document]
kirillt: Also, do I understand correctly that we should not store cache after running branch builds because it can be incompatible with other branches? Am I right that we still should save cache when a new commit is merged into `main`?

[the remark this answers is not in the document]
oluiscabral: @kirillt I'm not 100% sure what you mean by `because it can be incompatible with other branches`, but I see no problem storing branch build cache, considering that according to their README (https://github.com/Swatinem/rust-cache) unrelated branches don't affect each other:
image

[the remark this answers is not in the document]
Pushkarm029: > where the cache will be really stored in this case?

The cache is stored on GitHub servers. You can view the cached entries by navigating to the Cache section on the left-hand side of the Actions tab in this repository.

> because it can be incompatible with other branches?

No, that is not possible. You can read more about this [here](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache).

I recommended storing the cache on the `main` branch because:
- It prevents potential cache poisoning from experimental dependencies in (same) PR.
- Multiple PRs can benefit from a shared cache built from the main branch.

If each PR created its own cache, it would result in duplicate caches for common dependencies, leading to faster exhaustion of GitHub Actions storage.

>  Am I right that we still should save cache when a new commit is merged into main?


Yes, saving the cache when a new commit is merged into main is more efficient than doing so on a weekly basis.

.github/workflows/build.yml
Pushkarm029: I think so.

Pushkarm029:
@oluiscabral Let me know when you are ready for the final review.

oluiscabral:
> @oluiscabral Let me know when you are ready for the final review.

Yup, feel free to review it again whenever you want

.github/workflows/weekly.yml
```diff
@@ -19,6 +19,11 @@ jobs:
       - name: Install Rust
         uses: dtolnay/rust-toolchain@stable
 
+      - name: Set up Cargo Cache
+        uses: Swatinem/rust-cache@v2
+        with:
+          save-if: true
+
```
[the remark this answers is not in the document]
oluiscabral: So let me see if I'm following it correctly. The idea is to keep the weekly job, but adding a new workflow that has
```
on:
  push:
    branches:
      - main
``` as predicate instead of a weekly schedule, right?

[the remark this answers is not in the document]
Comment (Pushkarm029) [reply .github/workflows/weekly.yml:26]: Yup, the Weekly will remain same as we have in `main` right now. Cache will be created on pushing or pull request to `main`.

```yml
on:
  push:
    branches:
      - main
  pull_request:
    branches:
      - main
```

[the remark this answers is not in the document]
oluiscabral: Perfect :+1: got it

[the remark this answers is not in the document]
Pushkarm029: I think caching on push or PR to the main branch will be sufficient. Therefore, we should remove caching from the weekly workflow.
```yml
        with:
          save-if: false
```

oluiscabral:
So just to let you guys know, @Pushkarm029  and @kirillt, after a thorough analysis with the team, this PR will be updated to transform the `ark-core/java` library into AAR instead of JAR.

I'll be pushing those updates next

kirillt:
Sounds good @oluiscabral, also consider moving bindings code into `ark-android` repo if it makes things easier. Ideally, `ark-core` repo should be free of Android code, but we can wait with it.

oluiscabral:
> So just to let you guys know, @Pushkarm029 and @kirillt, after a thorough analysis with the team, this PR will be updated to transform the `ark-core/java` library into AAR instead of JAR.
> 
> I'll be pushing those updates next

Alright, @Pushkarm029 and @kirillt, now `ark-core/java` is an AAR! :tada:

The published package (AAR) is currently being referenced at https://github.com/ARK-Builders/ark-android/pull/110

.github/workflows/build.yml
```diff
@@ -21,20 +21,11 @@ jobs:
...
-      - name: Check
-        run: cargo check
 
-      - name: Format
-        run: |
-          cargo fmt --all -- --check
-          cargo clippy --workspace --bins -- -D warnings
-
-      - name: Build Debug
-        run: cargo build --verbose
```
oluiscabral, line 37: @Pushkarm029 @kirillt I removed those steps for now, because the behavior of the `Format` was not being 100% predictable. Since I started this PR, around a month ago, the `Format` step changed its results multiple times, even though the Rust code remained exactly the same

As example, you can check that from time to time some old and recent jobs didn't fully complete because formatting errors were randomly being thrown.

Most recent occurrence:
- https://github.com/ARK-Builders/ark-core/actions/runs/12120246957/job/33788525665 ran just fine
- https://github.com/ARK-Builders/ark-core/actions/runs/12287131386/job/34288596396 a week later started raising a formatting error

I judge as a bad practice using a `nightly` or any other "dynamic" version to format a project, since it can introduce errors from newly added formatting features :thinking: 

Curious to know what you guys think about it

kirillt: Interesting point. Actually, I prefer setting up `cargo fmt` everywhere, because it helps to avoid inconsistent code style (and standard Rust code style looks pretty fine). But I've never encountered `cargo fmt check` rejecting a bit outdated codebase.

**My guess** is that the code style used by `cargo fmt` develops including new decisions made by Rust community. And out code is modified too slowly so we don't include these new updates. If this is true, then we could keep `cargo fmt check` and actually reformat the code every time we encounter such failures on CI, but only manually.

@Pushkarm029 what do you think?

.github/workflows/build.yml
tareknaser, line 1: I think the clean approach here is to have two separate workflows: one for building and testing the Rust code, and another for handling the bindings, which would depend on the first workflow. I think we can use GitHub artifacts to pass the Rust binaries needed for the bindings.

Also, the benchmarking workflow should depend on the build/test workflow. I’ve noticed that PRs often get flooded with benchmarking comments when someone is fixing something in the CI, so clearly this setup isn’t ideal.

I think something like this could work
```yaml
name: Build Rust Code

on: ...

jobs:
  build:
    name: Build on ${{ matrix.os }}
    runs-on: ${{ matrix.os }}
    strategy:
      matrix:
        os: [ubuntu-latest, windows-latest, macos-14]
    steps:
      - uses: actions/checkout@v4

      - name: Install Rust
        uses: dtolnay/rust-toolchain@stable

      - name: Build Release
        run: cargo build --release

      - name: Upload Build Artifacts
        uses: actions/upload-artifact@v4
        with:
          name: build-artifacts-${{ matrix.os }}
          path: target/release/
```
and then
```yaml
name: Generate and Test Bindings

on:
  workflow_run:
    workflows: ["Build Rust Code"]
    types:
      - completed

jobs:
  bindings:
    name: Generate and Test Bindings on ${{ matrix.os }}
    runs-on: ${{ matrix.os }}
    strategy:
      matrix:
        os: [ubuntu-latest, windows-latest, macos-14]
    steps:
      - uses: actions/checkout@v4

      - name: Download Build Artifacts
        uses: actions/download-artifact@v4
        with:
          name: build-artifacts-${{ matrix.os }}
          path: target/release/

      # Java, Android SDK and NDK
```
Same for the benchmarking workflow 

But I don’t think this should block the PR any longer, especially since it’s affecting the Android team as well. The CI workflow isn’t terribly long at the moment, so let’s move forward with this and I can address it in a follow-up PR.

tareknaser, line 1: getting this warning locally, and I can also see it in the CI:
```sh
> Task :lib:compileReleaseJavaWithJavac
Java compiler version 22 has deprecated support for compiling with source/target version 8.
warning: [options] source value 8 is obsolete and will be removed in a future release
Try one of the following options:
    1. [Recommended] Use Java toolchain with a lower language version
    2. Set a higher source/target version
    3. Use a lower version of the JDK running the build (if you're not using Java toolchain)
For more details on how to configure these settings, see https://developer.android.com/build/jdks.
To suppress this warning, set android.javaCompile.suppressSourceTargetDeprecationWarning=true in gradle.properties.
warning: [options] target value 8 is obsolete and will be removed in a future release
warning: [options] To suppress warnings about obsolete options, use -Xlint:-options.
3 warnings
warning: [options] source value 8 is obsolete and will be removed in a future release
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,57 @@
...
+      contents: read
+      packages: write
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          targets: aarch64-linux-android armv7-linux-androideabi i686-linux-android x86_64-linux-android
+
```
tareknaser, line 21: Could you add these targets to the instructions in `java/README.md`, specifically the `rustup` command?

java/lib/build.gradle
```diff
@@ -0,0 +1,85 @@
+apply plugin: 'maven-publish'
+apply plugin: 'com.android.library'
+apply plugin: 'org.mozilla.rust-android-gradle.rust-android'
+
+android {
+    ndkVersion '28.0.12433566'
```
tareknaser, line 6: Latest version is `28.0.12674087`

.github/workflows/release.yml
```diff
@@ -0,0 +1,57 @@
+name: Release
+
+on:
+  push:
+    tags:
+      - "*"
```
tareknaser, line 6: We probably don’t want to release on every tag. Can we configure it to only trigger releases for tags that start with ‘v’, like ‘v1.0.0’? I think that’s the more standard way to set this up.
```suggestion
on:
  push:
    tags:
      - 'v*'
```

.github/workflows/cache.yml
tareknaser, line 1: Not sure if we really need a separate workflow just for the cache (`.github/workflows/cache.yml`). I see that it only runs on push/PR to the main branch. Can we modify the save condition in the main workflow to:
```sh
save-if: ${{ github.ref == 'refs/heads/main' }}
```
and remove the redundant workflow?

.github/workflows/build.yml
tareknaser: We need to use nightly for formatting because some of the rules we define in `rustfmt.toml` are only supported in the nightly version. Rust nightly is updated frequently, that’s not a problem — it’s the pre-release version. We have two options:
	•	We can abandon the rules we’ve defined (I would like to keep them)
	•	We can check for formatting in a separate job that runs nightly, within the same workflow.
For example:
```yaml
jobs:
  format:
    name: Code Formatting Check
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4

      - name: Install Nightly Rust with rustfmt
        uses: dtolnay/rust-toolchain@stable
        with:
          toolchain: nightly
          components: rustfmt

      - name: Run rustfmt
        run: cargo fmt --all -- --check

  build_and_test:
    name: Build and Test
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4

      - name: Install Stable Rust
        uses: dtolnay/rust-toolchain@stable
        with:
          toolchain: stable

      ...
```

java/lib/build.gradle
oluiscabral: Nice, I'll update it. At the time this PR was created, `28.0.12433566` was the latest version :disguised_face:

.github/workflows/build.yml
oluiscabral: Hmm I see, I'll investigate it

.github/workflows/release.yml
oluiscabral: Good idea, do you mean to add a step-by-step on how to locally cross-compile the library?

tareknaser: When I tried running this locally, it failed because I didn’t have these targets set up, and I had to debug it myself. It would be helpful to include a command for adding these targets as a perquisite, something like:
```sh
rustup target add \
    aarch64-linux-android \
    armv7-linux-androideabi \
    ...
```

tareknaser: Also, let’s include a note about the supported `ndkVersion`, because running this with a different version also causes it to completely fail locally for me (I had version `28.0.12674087`. This is why I suggested https://github.com/ARK-Builders/ark-core/pull/94#discussion_r1885367339)

with an annoying error as well (I think it said something like `ndk` not installed)

.github/workflows/build.yml
tareknaser: Thank you

java/lib/build.gradle
tareknaser: > At the time this PR was created, `28.0.12433566` was the latest version 🥸

I see the sarcasm, but let’s maintain a professional tone moving forward. Thank you.

oluiscabral: Oh sorry @tareknaser, it was not sarcasm, I was just stating that I didn't know a new version was released. Sorry if you felt that way. It was actually nice that you let me know my local NDK is outdated

.github/workflows/build.yml
kirillt: We actually care about formatting only during merging, so we probably can configure GitHub rules and prevent merge if `cargo fmt` job fails specifically. This would allow see that the build itself works.

.github/workflows/build.yml
kirillt: @tareknaser feel free to create new PR targeting this branch `patch-1`, which we could start reviewing right after this one is merged.

kirillt: Aaahh, that's why would be great to move all the Java code to `ark-android`.. especially after we switched from JAR to AAR, so we don't do "plain Java" anymore and really only use Java for Android..

.github/workflows/release.yml
kirillt: Good idea, also could be `v*.*.*` to allow only semantic versioned releases.

(Not 100% sure about the this mask)

.github/workflows/cache.yml
kirillt: @Pushkarm029

.github/workflows/build.yml
oluiscabral: Yes. Maybe we could even move the JNI Rust related code to `ark-android` then, since it is specifically related to Java as well

tareknaser: > @tareknaser feel free to create new PR targeting this branch patch-1, which we could start reviewing right after this one is merged.

https://github.com/ARK-Builders/ark-core/pull/96

opened a pull request against this one. It should be pretty simple. I tested it on my fork and included the links in the description so you can see it in action.

.github/workflows/cache.yml
tareknaser: Whether we keep this workflow or not, I don’t think it’s worth running it on macOS ARM (`macos-13-xlarge`) with every push to main. We only use the macOS ARM workflow in the weekly build since it’s a paid runner. 
Or, we could increase the retention period for this specific GitHub Actions artifact.

.github/workflows/build.yml
kirillt: Merged

.github/workflows/release.yml
kirillt: @tareknaser this one seems to be resolved, right?

.github/workflows/build.yml
kirillt: @oluiscabral exactly! do you think it's reasonable amount of work for the current pair of PRs, or should we merge these first, and proceed with refactoring in new PRs?

.github/workflows/build.yml
kirillt: @oluiscabral I think we don't have `cargo clippy --workspace --bins -- -D warnings` anymore..

Not sure if it's really any valuable though, does anyone looks into CI warnings? 🤔 

cc @tareknaser @Pushkarm029

Pushkarm029: i think `cargo clippy --workspace --bins -- -D warnings` is necessary, It helps us catch common mistakes, anti-patterns, and potential improvements that the compiler doesn't check for. 

So, I suggest we should add it between the `format` and `build-and-test` job.

.github/workflows/cache.yml
Pushkarm029: > and remove the redundant workflow?

Agreed

Pushkarm029 (approved):
Great work

@oluiscabral, Let's resolve the minor changes now and save the major ones for the next PR or create an issue for them. This PR is already quite large and has been open for months.

PR #95, opened 2024-11-17
`fs-cache`: Add Cache Struct

Pushkarm029:
- Cache is not implemented yet. It will be using `MemoryLimitedStorage` under the hood.
- MemoryLimitedStorage is a simplified version of FolderStorage that keeps a small number of (K, V) pairs in memory. 
- Currently, it uses Least Recently Used (LRU) algorithm with a LinkedHashMap to decide which (K, V) pairs to keep in memory.


fs-cache/src/cache.rs
```diff
@@ -0,0 +1,69 @@
...
+
+    /// Load most recent cached items into memory based on timestamps
+    pub fn load_recent(&mut self) -> Result<()> {
+        self.storage.load_fs()
+    }
+
+    /// Get number of items currently in memory
+    // pub fn memory_items(&self) -> usize {
+    //     self.storage.memory_items()
+    // }
```
kirillt, line 58: Should be good idea to log such information with `debug` level. Also, we should count bytes not number of entries. Some cache items can be huge, e.g. bitmaps.

fs-cache/src/cache.rs
```diff
@@ -0,0 +1,69 @@
...
+
+    /// Get sync status between memory and disk
+    pub fn sync_status(&self) -> Result<SyncStatus> {
+        self.storage.sync_status()
+    }
```
kirillt, line 63: We don't really need complicated sync. It's ok to always write new values to disk.

The only thing we should be careful about is overwriting. If a key already has some value attached to it, it must be same as what we attempt to write. We should skip overwriting. We can use `debug_assert!` to double-check that the values are same during development.

When we checked that the value is absent and we start writing, we must ensure atomic write to disk. If two apps on same devices write to same file, it can result in half-written value. Should we use [this method to ensure atomic writing](https://github.com/ARK-Builders/ark-core/blob/main/fs-atomic-light/src/lib.rs)?

fs-cache/src/cache.rs
```diff
@@ -0,0 +1,69 @@
...
+
+    /// Sync changes to disk
+    pub fn sync(&mut self) -> Result<()> {
+        self.storage.sync()
+    }
```
kirillt, line 68: We don't really need complicated sync. It's ok to always write new values to disk.

The only thing we should be careful about is overwriting. If a key already has some value attached to it, it must be same as what we attempt to write. We should skip overwriting. We can use `debug_assert!` to double-check that the values are same during development.

When we checked that the value is absent and we start writing, we must ensure atomic write to disk. If two apps on same devices write to same file, it can result in half-written value. Should we use [this method to ensure atomic writing](https://github.com/ARK-Builders/ark-core/blob/main/fs-atomic-light/src/lib.rs)?

fs-cache/src/cache.rs
tareknaser, line 1: Can we make use of [`lru`](https://crates.io/crates/lru) or another crate like [`uluru`](https://crates.io/crates/uluru)? `lru` seems to be actively maintained. 

I haven't looked into either of them yet, but it would be great to reuse some of the tested code from these crates directly and maybe add custom methods as needed.

fs-cache/src/cache.rs
```diff
@@ -0,0 +1,60 @@
+use crate::memory_limited_storage::MemoryLimitedStorage;
+use data_error::Result;
+use std::path::Path;
+
+pub struct Cache<K, V> {
+    storage: MemoryLimitedStorage<K, V>,
+}
```
tareknaser, line 7: `Cache` is just a very thin wrapper around `MemoryLimitedStorage`, to the point where it doesn't really add any new functionality. 

Why do we need it? Are there plans to extend it in the future?

fs-cache/src/memory_limited_storage.rs
```diff
@@ -0,0 +1,273 @@
...
+use std::time::SystemTime;
+
+use data_error::{ArklibError, Result};
+use linked_hash_map::LinkedHashMap;
+
+pub struct MemoryLimitedStorage<K, V> {
+    /// Label for logging
+    label: String,
+    /// Path to the underlying folder where data is persisted
+    path: PathBuf,
```
tareknaser, line 13: Why do we need a separate folder for persistence (where each key-value pair stored as a different file)? Can't we just use a single file? 

If each value can be serialized to a string, we could store all values in a single JSON file.

fs-cache/src/memory_limited_storage.rs
```diff
@@ -0,0 +1,273 @@
...
+        let mut file = File::create(&file_path)?;
+        file.write_all(serde_json::to_string_pretty(&value)?.as_bytes())?;
+        file.flush()?;
+
+        let new_timestamp = SystemTime::now();
+        file.set_modified(new_timestamp)?;
+        file.sync_all()?;
+
+        Ok(())
+    }
```
tareknaser, line 194: Some of the code here is very similar to what we have in `FolderStorage`.

This makes me wonder if we can abstract it away by using `FileStorage`/`FolderStorage` from `fs-storage` crate? I think this would make a cleaner structure.

fs-cache/src/memory_limited_storage.rs
tareknaser, line 1: Could you write doc comments for the public API? Specifically, for the public structs and methods?

tareknaser, line 1: Could you add tests for the new functionality?

Cargo.toml
tareknaser, line 1: We should probably add an entry to [this section](https://github.com/ARK-Builders/ark-core?tab=readme-ov-file#crates)? It could mention that it's still experimental.

fs-cache/src/memory_limited_storage.rs
```diff
@@ -0,0 +1,273 @@
...
+            let path = entry.path();
+            if path.is_file()
+                && path
+                    .extension()
+                    .map_or(false, |ext| ext == "json")
+            {
+                if let Ok(metadata) = fs::metadata(&path) {
+                    let key = extract_key_from_file_path(&self.label, &path)?;
+                    file_metadata.push((key, metadata.len() as usize));
+                }
```
tareknaser, line 123: This assumes that every file with valid JSON is considered a valid key. Is this an intentional assumption?

fs-cache/src/cache.rs
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    pub fn get(&mut self, key: &K) -> Option<V> {
+        let result = self.storage.get(key);
+        log::debug!(
+            "{} cache: get key={} -> found={}",
+            self.storage.label(),
+            key,
+            result.is_some()
+        );
```
tareknaser, line 42: `result.is_some()` returns true if the option is a `Some` value, so printing that isn’t very useful. We could just print `result` directly.

fs-cache/src/cache.rs
```diff
@@ -0,0 +1,60 @@
...
+        if self.storage.get(&key).is_some() {
+            log::debug!(
+                "{} cache: skip setting existing key={}",
+                self.storage.label(),
+                key
+            );
+            return Ok(());
+        }
+
+        log::debug!("{} cache: set key={}", self.storage.label(), key);
```
tareknaser, line 57: This debug statement is misleading. It should be placed after the call to `self.storage.set()`

fs-cache/src/memory_limited_storage.rs
```diff
@@ -0,0 +1,273 @@
...
+                value_size,
+                self.max_memory_bytes
+            );
+            return;
+        }
+
+        // Remove oldest entries until we have space for new value
+        while self.current_memory_bytes + value_size > self.max_memory_bytes
+            && !self.memory_cache.is_empty()
+        {
```
tareknaser, line 226: I think `!self.memory_cache.is_empty()` is redundant. We already checked that `value_size` is less than `self.max_memory_bytes`, so it's guaranteed that the value will fit in the cache if the cache is empty. 

Is there a case I might be missing?

fs-cache/src/cache.rs
```diff
@@ -0,0 +1,60 @@
...
+            "{} cache: get key={} -> found={}",
+            self.storage.label(),
+            key,
+            result.is_some()
+        );
+        result
+    }
+
+    pub fn set(&mut self, key: K, value: V) -> Result<()> {
+        // Check if value already exists
```
tareknaser, line 47: What happens if a value already exists? This behavior isn’t documented.

fs-cache/src/memory_limited_storage.rs
```diff
@@ -0,0 +1,273 @@
...
+        );
+
+        Ok(())
+    }
+
+    fn estimate_size(value: &V) -> usize {
+        serde_json::to_vec(value)
+            .map(|v| v.len())
+            .unwrap_or(0)
+    }
```
tareknaser, line 180: `estimate_size` returns 0 on failure.

fs-cache/src/memory_limited_storage.rs
```diff
@@ -0,0 +1,273 @@
...
+            ArklibError::Storage(
+                self.label.clone(),
+                format!("Failed to read value for key {}: {}", key, err),
+            )
+        })?;
+        Ok(value)
+    }
+
+    fn add_to_memory_cache(&mut self, key: K, value: V) {
+        let value_size = Self::estimate_size(&value);
```
tareknaser, line 210: I think this breaks the logic of `add_to_memory_cache`.

fs-cache/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+data-error = { path = "../data-error" }
+data-resource = { path = "../data-resource" }
+fs-storage = { path = "../fs-storage"}
+linked-hash-map = "0.5.6"
+
+[dev-dependencies]
+anyhow = "1.0.81"
+quickcheck = { version = "1.0.3", features = ["use_logging"] }
+quickcheck_macros = "1.0.0"
+tempdir = "0.3.7"
```
tareknaser, line 24: None of the `dev-dependencies` are actually used.

fs-cache/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+name = "fs_cache"
+crate-type = ["rlib", "cdylib"]
+bench = false
+
+[dependencies]
+log = { version = "0.4.17", features = ["release_max_level_off"] }
+serde_json = "1.0.82"
+serde = { version = "1.0.138", features = ["derive"] }
+data-error = { path = "../data-error" }
+data-resource = { path = "../data-resource" }
```
tareknaser, line 16: `data-resource` never used

fs-cache/src/memory_limited_storage.rs
```diff
@@ -0,0 +1,273 @@
...
+use linked_hash_map::LinkedHashMap;
+
+pub struct MemoryLimitedStorage<K, V> {
+    /// Label for logging
+    label: String,
+    /// Path to the underlying folder where data is persisted
+    path: PathBuf,
+    /// In-memory LRU cache combining map and queue functionality
+    memory_cache: LinkedHashMap<K, V>,
+    // Bytes present in memory
```
tareknaser, line 16: Nit: This should be a doc comment.

fs-cache/src/memory_limited_storage.rs
```diff
@@ -0,0 +1,273 @@
...
+use data_error::{ArklibError, Result};
+use linked_hash_map::LinkedHashMap;
+
+pub struct MemoryLimitedStorage<K, V> {
+    /// Label for logging
+    label: String,
+    /// Path to the underlying folder where data is persisted
+    path: PathBuf,
+    /// In-memory LRU cache combining map and queue functionality
+    memory_cache: LinkedHashMap<K, V>,
```
tareknaser, line 15: Nit: I think this comment is slightly incorrect. `LinkedHashMap` is not the cache itself.

fs-cache/src/cache.rs
Pushkarm029: Initially, I implemented a few functions specifically for `Cache`, but Kirill suggested that we currently only need the `get`/`set` API. Therefore, I think it's a good idea to remove the wrapper and directly expose this API.

Pushkarm029:
Thank you for the review.

ark-cli/src/util.rs
```diff
@@ -179,10 +179,9 @@ pub fn translate_storage(
     root: &Option<PathBuf>,
     storage: &str,
 ) -> Option<(PathBuf, Option<StorageType>)> {
-    if let Ok(path) = PathBuf::from_str(storage) {
-        if path.exists() && path.is_dir() {
-            return Some((path, None));
-        }
+    let Ok(path) = PathBuf::from_str(storage);
```
kirillt, line 182: Does it panic if `from_str` returns `Err`?

Cargo.toml
kirillt: Right, that's a good idea.

fs-cache/src/cache.rs
kirillt: Actually, values can be huge. Like 100KB image. We should not log them. We can add parameter `log_values` to the storage constructor though, so small values would be logged all the time (responsibility of app developer).

I would refactor logging here like this:
- `log::debug("cache/{}: retrieving value for key {}", ...)`
- `self.storage.get`
- `log::warn` if the result is `None`

fs-cache/src/cache.rs
kirillt: Good catch. It can be indicator of redundant computation.

Since the cache uses `FolderStorage` under the hood, checking for key existence should be faster than retrieving the whole value.. Am I right that just checking the file existence is faster than reading the file?

I suggest the following:
- return error when the entry already exists
- add `keys` query to the API, that returns all known keys
- add `exists` predicate to the API

Query `keys` could be used by app developers to efficiently filter already known values, so they could skip not only writing to the disk, but entire redundant computation.

fs-cache/src/cache.rs
kirillt: @tareknaser I think it's better to put the trace before the action. If something crashes down the stack, we'll have the log telling us what we tried to do before the crash.

fs-cache/src/memory_limited_storage.rs
kirillt: @tareknaser `FolderStorage` is primary storage for cache, because values can be large, e.g. previews.

With time, it would be great to also implement cache backed by `FileStorage`. And sometime we're going to implement `ChunkedStorage` as well (hybrid between the two flavours).

fs-cache/src/memory_limited_storage.rs
```diff
@@ -0,0 +1,273 @@
...
+            self.label,
+            self.current_memory_bytes,
+            total_bytes
+        );
+
+        Ok(())
+    }
+
+    fn estimate_size(value: &V) -> usize {
+        serde_json::to_vec(value)
```
kirillt, line 177: We should avoid this. It should be possible to return number of bytes read from `load_value_from_disk`.

fs-cache/src/memory_limited_storage.rs
```diff
@@ -0,0 +1,273 @@
+use std::fs::{self, File};
+use std::io::Write;
+use std::path::{Path, PathBuf};
+use std::time::SystemTime;
+
+use data_error::{ArklibError, Result};
+use linked_hash_map::LinkedHashMap;
+
+pub struct MemoryLimitedStorage<K, V> {
```
kirillt, line 9: It seems that we can safely merge `Cache` and `MemoryLimitedStorage` into a single struct unless we find other use-case for `MemoryLimitedStorage` except caching.

fs-cache/src/memory_limited_storage.rs
```diff
@@ -0,0 +1,273 @@
...
+        Ok(storage)
+    }
+
+    pub fn label(&self) -> String {
+        self.label.clone()
+    }
+
+    pub fn get(&mut self, key: &K) -> Option<V> {
+        // Check memory cache first - will update LRU order automatically
+        if let Some(value) = self.memory_cache.get_refresh(key) {
```
kirillt, line 57: That's pretty cool, I didn't know about `get_refresh` function.

But I guess it works with _O(N)_ complexity, right? Consider using dedicated Rust crate for LRU to get _O(1)_ complexity. This isn't pressing issue, we can create an issue for it.

fs-cache/src/memory_limited_storage.rs
kirillt: We can try, but `FolderStorage` contains some logic for handling sync conflicts because it was implemented with user data in mind. For cache, we have simpler scenario because conflicts never arise thanks to deterministic computations (thanks to content addressing).

Let's keep this storage simple. We can build smarter modular crates later and use this simple crate as reference in unit tests.

kirillt: Some more context, comparing to `fs-cache`, `fs-storage` has more moving parts: versioning and conflicts resolution. It's not clear yet if we should merge `fs-atomic-versions` into `fs-storage`, or the opposite, extract `fs-monoid` from `fs-storage`. Modular approach is preferred but we can bog down in it, so I would postpone it for future versions.

fs-cache/src/memory_limited_storage.rs
```diff
@@ -0,0 +1,273 @@
...
+            return Some(value.clone());
+        }
+
+        // Try to load from disk
+        let file_path = self.path.join(format!("{}.json", key));
+        if file_path.exists() {
+            // Doubt: Update file's modiied time (in disk) on read to preserve LRU across app restarts?
```
kirillt, line 64: Let's track this feature and work on it later. Better to keep implementation simple for this moment and avoid redundant state. Btw we can also simply write cached keys into a file + apply atomic versioning on it, so all peers would have same view of LRU.

fs-cache/src/memory_limited_storage.rs
```diff
@@ -0,0 +1,273 @@
...
+    // Write a single value to disk
+    fn write_value_to_disk(&mut self, key: &K, value: &V) -> Result<()> {
+        let file_path = self.path.join(format!("{}.json", key));
+        let mut file = File::create(&file_path)?;
+        file.write_all(serde_json::to_string_pretty(&value)?.as_bytes())?;
+        file.flush()?;
+
+        let new_timestamp = SystemTime::now();
+        file.set_modified(new_timestamp)?;
+        file.sync_all()?;
```
kirillt, line 191: Do we really need it?

fs-cache/src/memory_limited_storage.rs
```diff
@@ -0,0 +1,273 @@
...
+    fn estimate_size(value: &V) -> usize {
+        serde_json::to_vec(value)
+            .map(|v| v.len())
+            .unwrap_or(0)
+    }
+
+    // Write a single value to disk
+    fn write_value_to_disk(&mut self, key: &K, value: &V) -> Result<()> {
+        let file_path = self.path.join(format!("{}.json", key));
+        let mut file = File::create(&file_path)?;
```
kirillt, line 185: Let's add `debug_assert` that the file doesn't exist.

kirillt: Also, we should use [lightweight atomic writing](https://github.com/ARK-Builders/ark-core/blob/main/fs-atomic-light/src/lib.rs) to avoid dirty writing. Keep in mind scenario when several ARK apps on same device use same folder and write to the cache in parallel.

I believe, that [atomic versions](https://github.com/ARK-Builders/ark-core/tree/main/fs-atomic-versions) would be excessive here, but I'm not 100% sure yet.

fs-cache/src/memory_limited_storage.rs
```diff
@@ -0,0 +1,273 @@
...
+        file.sync_all()?;
+
+        Ok(())
+    }
+
+    // Load a single value from disk
+    fn load_value_from_disk(&self, key: &K) -> Result<V> {
+        let file_path = self.path.join(format!("{}.json", key));
+        let file = File::open(&file_path)?;
+        let value: V = serde_json::from_reader(file).map_err(|err| {
```
kirillt, line 200: It's not necessary JSON, can be any binary (e.g. JPG).

App developer will decide the structure of the value, we should not have any assumptions around it.

fs-cache/src/memory_limited_storage.rs
kirillt: Agree, we should simply take number of bytes that were actually read from disk.

fs-cache/src/memory_limited_storage.rs
kirillt: Good catch.

1. If `self.memory_cache.is_empty()` is true, then `self.current_memory_bytes == 0`.
2. Then `self.current_memory_bytes + value_size > self.max_memory_bytes` is same as `value_size > self.max_memory_bytes`.
3. But `value_size > self.max_memory_bytes` has just been checked.

fs-cache/src/memory_limited_storage.rs
```diff
@@ -0,0 +1,273 @@
...
+        }
+
+        // Remove oldest entries until we have space for new value
+        while self.current_memory_bytes + value_size > self.max_memory_bytes
+            && !self.memory_cache.is_empty()
+        {
+            if let Some((_, old_value)) = self.memory_cache.pop_front() {
+                self.current_memory_bytes = self
+                    .current_memory_bytes
+                    .saturating_sub(Self::estimate_size(&old_value));
```
kirillt, line 230: Now I understand, why you've defined this function. But I think we can store sizes in the in-memory mapping next to the values themselves. That would be precise and faster. We can consider the memory overhead as fair trade-off.

kirillt: I would also add `debug_assert` that `current_memory_bytes  - old_value >= 0`.

fs-cache/src/memory_limited_storage.rs
```diff
@@ -0,0 +1,273 @@
...
+            self.max_memory_bytes
+        );
+    }
+}
+
+fn extract_key_from_file_path<K>(label: &str, path: &Path) -> Result<K>
+where
+    K: std::str::FromStr,
+{
+    path.file_stem()
```
kirillt, line 252: Let's simply use the whole filename as the key, no need for extension.

fs-cache/src/memory_limited_storage.rs
tareknaser: > because values can be large, e.g. previews

Right, that makes sense.

fs-cache/src/memory_limited_storage.rs
tareknaser: Sounds good. I agree, but I’d also love to see the core logic extracted from `fs-storage` in a later version and reused here if possible.

> fs-storage has more moving parts: versioning and conflict resolution

For example, this should ultimately be optional in one way or another. I’m not yet sure how it should be structured to create a clean API, and I think this is beyond the scope of this PR.

Marking this as resolved for now

fs-cache/src/cache.rs
tareknaser: That's fine. I don’t have a strong opinion on this, but I think it should be something like 'trying to set key "{}"' since right now it sounds like the value is already set when it’s not. Maybe I’m overthinking it.

fs-cache/src/memory_limited_storage.rs
Pushkarm029: No, not needed anymore, since we are no longer doing any kind of synchronization.

ark-cli/src/util.rs
Pushkarm029: - Even though PathBuf::from_str() has a return type of Result<Self, Self::Err>, it never fails because a PathBuf is just a cross-platform wrapper around a string path.

- The actual validation of whether the path exists or is valid in the filesystem happens later when you call methods like exists() or is_dir().

```rust
#[stable(feature = "path_from_str", since = "1.32.0")]
impl FromStr for PathBuf {
    type Err = core::convert::Infallible;

    #[inline]
    fn from_str(s: &str) -> Result<Self, Self::Err> {
        Ok(PathBuf::from(s))
    }
}
```

fs-cache/src/cache.rs
Pushkarm029: Removed it

fs-cache/src/memory_limited_storage.rs
Pushkarm029: Yes, initially, I had made this assumption. However, since our main objective for this cache is to store large values like bitmaps, images, etc., it is no longer needed. Thank you for pointing it out.

fs-cache/src/cache.rs
Pushkarm029: Added!

fs-cache/src/cache.rs
Pushkarm029: I think logging both (before setting and after) will be more beneficial. I have added it.

fs-cache/src/memory_limited_storage.rs
Pushkarm029: Added

fs-cache/src/memory_limited_storage.rs
Pushkarm029: I think we may need it, in `fn load_fs`, we compare timestamps to load recent files into memory. This could help us efficiently decide which files to load or skip since it provides precise timing. However, it's a very minor improvement, so we can skip it.

fs-cache/src/cache.rs
```diff
@@ -0,0 +1,668 @@
...
+        log_values: bool,
+    ) -> Result<Self> {
+        let mut cache = Self {
+            label: label.clone(),
+            path: PathBuf::from(path),
+            // TODO: NEED FIX
+            memory_cache: LruCache::new(
+                NonZeroUsize::new(max_memory_bytes)
+                    .expect("Capacity can't be zero"),
+            ),
```
Pushkarm029, line 68: LruCache requires the capacity (number of items) to be specified during initialization. However, our Cache is designed to be limited by max_memory_bytes. So, my question is: what would be the most efficient parameter to initialize the LruCache with? In all other functions, we are already comparing based on the number of bytes, not the number of items.

I think we can create another parameter which will be `Option<usize>` with default as 100 o maybe 1000. just to initialize this LruCache everything else is fine.

fs-cache/src/cache.rs
```diff
@@ -0,0 +1,668 @@
...
+
+        Ok(())
+    }
+
+    /// Retrieves a value from the memory cache.
+    fn fetch_from_memory(&mut self, key: &K) -> Option<V> {
+        self.memory_cache
+            .get(key)
+            .map(|entry| entry.value.clone())
+    }
```
tareknaser, line 249: If we remove `CacheEntry` (essentially removing the extra field for size) and use `V` directly, we can simply call `self.memory_cache.get(key)`. This would avoid cloning during each get operation, which I think can add up.

fs-cache/src/cache.rs
```diff
@@ -0,0 +1,668 @@
...
+                );
+                None
+            }
+        }
+    }
+
+    /// Writes a value to disk using atomic operations.
+    fn persist_to_disk(&mut self, key: &K, value: &V) -> Result<()> {
+        log::debug!("cache/{}: writing to disk for key {}", self.label, key);
+        fs::create_dir_all(&self.path)?;
```
tareknaser, line 287: I think we should return an error if the cache directory doesn’t exist for any reason. We create the directory when `new` is called (via `load_fs`), so it should always be there. If it’s missing, that’s not a valid state, and we should trigger an error.

fs-cache/src/cache.rs
```diff
@@ -0,0 +1,668 @@
...
+    }
+
+    /// Checks if a value exists either in memory or on disk.
+    pub fn exists(&self, key: &K) -> bool {
+        self.memory_cache.contains(key)
+            || self.path.join(key.to_string()).exists()
+    }
+
+    /// Returns an ordered iterator over all cached keys.
+    pub fn keys(&self) -> Result<impl Iterator<Item = K>> {
```
tareknaser, line 146: How is the iterator ordered? Is it by key, by insertion order, or something else? I don’t see any specific ordering in the code.

fs-cache/src/cache.rs
```diff
@@ -0,0 +1,668 @@
...
+    pub fn set(&mut self, key: K, value: V) -> Result<()> {
+        log::debug!("cache/{}: setting value for key {}", self.label, key);
+
+        // Check if value already exists
+        if self.exists(&key) {
+            return Err(ArklibError::Storage(
+                self.label.clone(),
+                format!("Key {} already exists in cache", key),
+            ));
+        }
```
tareknaser, line 127: I think it’s an intuitive feature for a `set` operation to overwrite the value if it already exists. It might be better if the developer at least had the option to decide whether to overwrite the value or not.

fs-cache/src/cache.rs
```diff
@@ -0,0 +1,668 @@
...
+    fn persist_to_disk(&mut self, key: &K, value: &V) -> Result<()> {
+        log::debug!("cache/{}: writing to disk for key {}", self.label, key);
+        fs::create_dir_all(&self.path)?;
+
+        let file_path = self.path.join(key.to_string());
+        debug_assert!(
+            !file_path.exists(),
+            "File {} should not exist before writing",
+            file_path.display()
+        );
```
tareknaser, line 294: Same here

fs-cache/src/cache.rs
```diff
@@ -0,0 +1,668 @@
...
+    /// Path to the underlying folder where data is persisted
+    path: PathBuf,
+    /// An in-memory LRU cache for quick access to frequently used items.
+    memory_cache: LruCache<K, CacheEntry<V>>,
+    /// The current memory usage in bytes.
+    current_memory_bytes: usize,
+    /// The maximum allowable memory usage in bytes.
+    max_memory_bytes: usize,
+    /// Whether to include values in debug logs
+    log_values: bool,
```
tareknaser, line 35: The `log_values` field is only read once in the code (on line 95), and I see logging occurring without checking it in other parts. Is this particular log statement somehow different? 

I think we can safely assume it’s always enabled and remove this field entirely.

fs-cache/src/cache.rs
```diff
@@ -0,0 +1,668 @@
...
+    }
+
+    /// Retrieves a value by its key, checking memory first then disk.
+    /// Returns None if the key doesn't exist.
+    pub fn get(&mut self, key: &K) -> Option<V> {
+        log::debug!("cache/{}: retrieving value for key {}", self.label, key);
+
+        let value = self
+            .fetch_from_memory(key)
+            .or_else(|| self.fetch_from_disk(key));
```
tareknaser, line 91: Here, I think this should be refactored to log whether the values were retrieved from memory or disk (if found).

fs-cache/src/cache.rs
```diff
@@ -0,0 +1,668 @@
...
+    use std::{
+        fs::File,
+        io::Write,
+        time::{Duration, SystemTime},
+    };
+    use tempdir::TempDir;
+
+    // Helper struct that implements required traits
+    #[derive(Clone, Debug, PartialEq)]
+    struct TestValue(Vec<u8>);
```
tareknaser, line 390: I’m fairly confident we can use `Vec<u8>` directly instead of creating a new struct like `TestValue`. I tried it locally and it works, so I think it’s better to go with `Vec<u8>` directly.

fs-storage/src/utils.rs
```diff
@@ -51,6 +53,56 @@ where
     Ok(data)
 }
 
+/// Writes a serializable value to a file and returns the timestamp of the write
+pub fn write_json_file<T: Serialize>(
+    path: &Path,
+    value: &T,
+    time: SystemTime,
+) -> Result<()> {
```
tareknaser, line 61: I think accepting a timestamp in this function could lead to confusion. I checked all its usages, and it seems we’re always using `Time::now()` to get the timestamp. 

I think it would be better to remove the timestamp parameter and just call `Time::now()` directly inside the function.

fs-storage/src/utils.rs
```diff
@@ -51,6 +53,56 @@ where
     Ok(data)
 }
 
+/// Writes a serializable value to a file and returns the timestamp of the write
+pub fn write_json_file<T: Serialize>(
```
tareknaser, line 57: This function doesn't really return the timestamp of the write.

fs-storage/src/utils.rs
```diff
@@ -51,6 +53,56 @@ where
...
+    file.sync_all()?;
+
+    Ok(())
+}
+
+pub fn extract_key_from_file_path<K>(
+    label: &str,
+    path: &Path,
+    include_extension: bool,
+) -> Result<K>
```
tareknaser, line 76: Thanks for moving this function to the utils module. It definitely feels like the right place for it.

Could you also add a doc comment for it?

fs-cache/src/lib.rs
```diff
@@ -0,0 +1 @@
+pub mod cache;
```
tareknaser, line 1: All of the following comments are just nitpicks, so feel free to ignore them.

I think it’s better to only export the `Cache` struct. That way, when we import the crate, we can use `Cache` directly without needing to do `fs_cache::cache::Cache`. It would be `fs_cache::Cache` instead.


```suggestion
mod cache;

pub use cache::Cache;
```

fs-cache/src/cache.rs
```diff
@@ -0,0 +1,668 @@
...
+    #[fixture]
+    fn cache(temp_dir: TempDir) -> Cache<String, TestValue> {
+        Cache::new(
+            "test".to_string(),
+            temp_dir.path(),
+            1024 * 1024, // 1MB
+            false,
+        )
+        .expect("Failed to create cache")
+    }
```
tareknaser, line 418: I feel like using `fixtures` here might be a bit overkill, but maybe you have plans to expand on this in the future. Personally, I think we could just create the `temp_dir` and `cache` instances directly in each test function, or use a helper function, which would make the code more readable.

fs-cache/src/cache.rs
```diff
@@ -0,0 +1,668 @@
...
+        }
+
+        // Add new value and update size
+        self.memory_cache.put(
+            key.clone(),
+            CacheEntry {
+                value: value.clone(),
+                size,
+            },
+        );
```
tareknaser, line 362: A small suggestion: we could use `push` here instead of `put`. According to the docs, `push` does the following: ‘If an entry with key k already exists in the cache or another cache entry is removed (due to the LRU’s capacity), then it returns the old entry’s key-value pair.’

It should be a nice touch to log the evicted key, if there is one.

fs-cache/src/cache.rs
tareknaser, line 294: I don’t see `file_path` being used anywhere else, except in the `debug_assert`. If that’s the case, we can move it inside the `debug_assert` block.

README.md
```diff
@@ -28,18 +28,19 @@ The core crate is `fs-index` which provides us with [content addressing](https:/
...
-| `fs-storage`    | Filesystem storage for resources         |
-| `fs-metadata`   | Metadata management                      |
-| `fs-properties` | Properties management                    |
-| `data-link`     | Linking resources                        |
-| `data-pdf`      | PDF handling                             |
-| `data-error`    | Error handling                           |
-| `data-json`     | JSON serialization and deserialization   |
+| Package         | Description                               |
+| --------------- | ----------------------------------------  |
+| `ark-cli`       | The CLI tool to interact with ark crates  |
```
kirillt, line 33:
```suggestion
| `ark-cli`       | The CLI tool to interact with ARK crates  |
```

README.md
```diff
@@ -28,18 +28,19 @@ The core crate is `fs-index` which provides us with [content addressing](https:/
...
+| `data-resource` | Resource hashing and ID construction      |
+| `fs-cache`      | Memory and disk caching with LRU eviction |
+| `fs-index`      | Resource Index construction and updating  |
+| `fs-storage`    | Filesystem storage for resources          |
```
kirillt, line 37:
```suggestion
| `fs-storage`    | Key-value storage persisted on filesystem |
```

fs-cache/src/cache.rs
```diff
@@ -0,0 +1,668 @@
...
+use lru::LruCache;
+
+/// A cache entry that stores a value and its size in bytes.
+///
+/// This structure is used to track both the actual data (value)
+/// and its memory usage (size) in the cache.
+struct CacheEntry<V> {
+    value: V,
+    size: usize,
+}
```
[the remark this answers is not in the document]
kirillt: We need to track memory consumption in bytes to be precise about when to offload values, and to support large values, too. We probably can't avoid saving value sizes into memory, otherwise when we hit the limit we cannot know how much values we need to offload.

However, that's where we could split the crate into 2 flavours:
1. dynamically-sized values e.g. byte vectors and text strings
2. statically-sized values e.g. integers

For the 2nd flavour we could utilize some standard Rust trait.

Is there a way to have these 2 flavours combined nicely in a single crate?

fs-cache/src/cache.rs
kirillt: Guys, what about this? https://docs.rs/lru-mem/latest/lru_mem/

But it has only 3 stars on GitHub..

fs-cache/src/cache.rs
```diff
@@ -1,60 +1,489 @@
...
+        }
+
+        // Sort by size before loading
+        file_metadata.sort_by(|a, b| b.1.cmp(&a.1));
+
+        // Clear existing cache
+        self.memory_cache.clear();
+        self.current_memory_bytes = 0;
+
+        // Load files that fit in memory
```
kirillt: Interesting approach with pre-loading the cache. We probably should have a switch, so the app developer could turn it on/off.

fs-cache/src/cache.rs
kirillt: Here looks fine, because new key-value pairs could appear.

fs-cache/src/cache.rs
```diff
@@ -170,45 +175,42 @@ where
...
+                let metadata = entry.metadata()?;
+                let modified = metadata.modified()?;
+                let size = metadata.len() as usize;
+                file_metadata.push((key, size, modified));
             }
         }
 
-        // Sort by size before loading
-        file_metadata.sort_by(|a, b| b.1.cmp(&a.1));
+        // Sort by modified time (most recent first)
```
kirillt, line 198: Actually, I'm not sure that pre-loading the most recently modified values would really be beneficial.

We could implement more sophisticated approach with gathering query statistics and recording it somewhere on disk for pre-loading in future. But I would do it in a separate PR and not right now.

fs-cache/src/cache.rs
tareknaser: It's actually an interesting option and would have been a perfect fit 😃

I wouldn’t recommend it though because if we find any issues later in the crate, we’d have to fork it and fix the problem ourselves, and we’re not familiar with the code. Plus, since it's not actively maintained/ used, there wouldn’t be anyone around to help us either.

fs-cache/src/cache.rs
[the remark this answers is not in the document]
tareknaser: > We need to track memory consumption in bytes to be precise about when to offload values...

Thtat's fine but we're actually reading the file size from the disk again in the `get_file_size` method, even though we already have it stored in the metadata. That seems a bit wasteful. Check out the next comment for more on this

> 1. dynamically-sized values e.g. byte vectors and text strings

That's a good point
I completely missed the dynamic types aspect when I looked at this. Now it makes a lot more sense why we need to track the data size instead of just the number of items.


> Is there a way to have these 2 flavours combined nicely in a single crate?

If we're dealing with types that have a fixed size, like `usize`, it doesn't really matter if we count how many items there are or just the total size they take up. But this completely breaks with the second flavour you mentioned.

The only solution I can think of right now is to treat all types of data as if they were the second type – basically, keep track of how much memory they use instead of how many there are. But that brings up the question of how to do this in a clean way

Main thread: https://github.com/ARK-Builders/ark-core/pull/95#discussion_r1864880931

fs-cache/src/cache.rs
Pushkarm029: Done!

fs-cache/src/cache.rs
Pushkarm029: Done!

fs-storage/src/utils.rs
Pushkarm029: Tried doing it, but it failed some tests due to sync issues in both `file_storage` and `folder_storage`.

fs-cache/src/cache.rs
```diff
@@ -0,0 +1,668 @@
...
+                .pop_lru()
+                .expect("Cache should have entries to evict");
+            debug_assert!(
+                self.current_memory_bytes >= old_entry.size,
+                "Memory tracking inconsistency detected"
+            );
+            self.current_memory_bytes = self
+                .current_memory_bytes
+                .saturating_sub(old_entry.size);
+        }
```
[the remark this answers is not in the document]
Pushkarm029: Tried replacing the existing logic with resize(). However, this operation is still not O(1) because lru.resize has a complexity of O(N).

Additionally, using resize requires extra calculations to determine the target_size for the LRU cache. Considering this overhead, I think the existing approach is better.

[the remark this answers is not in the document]
Pushkarm029: I will explore the pros & cons of implementing our own LRU cache.

fs-cache/src/cache.rs
Pushkarm029: Added!

PR #96, opened 2024-12-19
CI Workflow Refactor

tareknaser:
This PR refactors the CI workflows. The biggest change is moving Java bindings to a separate job that only runs if the "Build and Test" jobs pass.
Related discussion: https://github.com/ARK-Builders/ark-core/pull/94#discussion_r1885367107

#### What’s New:
- Java bindings now run as a separate job after all the build and test jobs are done.
- Weekly Testing workflow is simplified with a matrix so we don’t repeat the same steps for each platform.
- Benchmarks now run only if Java bindings pass and it’s a pull request.

#### Why This Approach?
- Benchmarks
	- Tried keeping benchmarks in a separate workflow using `workflow_run`, but ran into issues with getting the `branchName` and commenting on PRs.
- Java Bindings
	- Looked at passing artifacts between workflows, but GitHub Actions doesn’t support this natively yet, and [the hacks I found](https://stackoverflow.com/questions/64868918/how-to-download-artifact-release-asset-in-another-workflow) weren’t clean.
	- Settled on keeping everything in the same workflow since [GitHub makes it easy to share data between jobs](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/storing-and-sharing-data-from-a-workflow#passing-data-between-jobs-in-a-workflow) without extra setup.

#### Current CI Setup:
Screenshot 2024-12-20 at 1 39 27 AM

#### Examples from my fork:
- **Push to Main**: Benchmarks are skipped  
  [Run Example](https://github.com/tareknaser/ark-rust/actions/runs/12421950970)
- **Pull Request**: Everything runs as expected, including benchmarks  
  [Run Example](https://github.com/tareknaser/ark-rust/actions/runs/12421961103)


Pushkarm029 (approved):
Clean✨

.github/workflows/build.yml
```diff
@@ -27,75 +26,56 @@ jobs:
...
-  linux:
-    name: Build on Linux
-    runs-on: ubuntu-latest
+  build-and-test:
+    name: Build and Test
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        # We also run on macOS arm in the weekly workflow.
+        os: [ubuntu-latest, windows-latest, macos-14]
```
Pushkarm029, line 35: We are excluding `MacOS ARM` here to avoid the cost on every commit?

tareknaser: Exactly

PR #98, opened 2025-01-31
`fs-cache`: Add LRU Struct

Pushkarm029:


fs-cache/src/lru.rs
```diff
@@ -0,0 +1,219 @@
...
+    bottom: Option<K>,
+}
+
+pub struct LRUCache<K, V> {
+    entries: HashMap<K, CacheEntry<K, V>>,
+    head: Option<K>,
+    tail: Option<K>,
+    memory_used: usize,
+    max_memory: usize,
+}
```
Pushkarm029, line 16: To remove few expensive clones, we can consider this later:
```
pub struct CacheEntry<V> {
    value: V,
    size: usize,
    top: Option<usize>,
    bottom: Option<usize>,
    id: usize,
}

pub struct LRUCache<K, V> {
    entries: HashMap<K, CacheEntry<V>>,
    keys: HashMap<usize, K>,
    head: Option<usize>,
    tail: Option<usize>,
    memory_used: usize,
    max_memory: usize,
    id_counter: usize,
}
```

PR #104, opened 2025-08-30
[ARK Drop CLI] Transfer Speedup

oluiscabral:


pushkarm029:
Awesome docs!

> Redundant help sub-command: Both drop-cli and drop-cli config have a help sub-command alongside the --help flag. The flag alone should be sufficient.

> Avatar flags: The --avatar and --avatar-b64 flags in the CLI seem unclear in purpose. If they're not essential for core functionality, consider removing them to simplify the interface.

These are not fixed yet.

oluiscabral:
> Redundant help sub-command: Both drop-cli and drop-cli config have a help sub-command alongside the --help flag. The flag alone should be sufficient.

This one didn't make sense to me. `drop-cli config` is another page. Just `drop-cli` is like "home page". So I don't see why shouldn't they have different help options. Can you elaborate?

pushkarm029:
> This one didn't make sense to me. drop-cli config is another page. Just drop-cli is like "home page". So I don't see why shouldn't they have different help options. Can you elaborate?

My bad, forget those pages. Let me clarify: in `drop-cli` there are currently two ways to run help: `drop-cli help` and `drop-cli --help`. I suggest we keep just one of them, either `drop-cli help` or `drop-cli --help`.

PR #105, opened 2025-09-03
[ARK Drop CLI] Enhancements

oluiscabral:


pushkarm029:
@oluiscabral

oluiscabral:
> image
> For me, when sending, it completes instantly right after ⏳ Waiting for receiver... (Press Ctrl+C to cancel). It doesn’t actually wait.

Interesting :thinking: 

What OS do you use? I'll take a better look

oluiscabral:
@pushkarm029

Regarding https://github.com/ARK-Builders/ark-core/pull/105#issuecomment-3262267916

Yeah, thanks for pointing out. I deleted the conditional that shuts the send router down. I'll put it right back

